PR #6, opened 2018-10-22
Update Searching.md

ScottPatterson:
Updating searching topic to include information about searching on types.

### DIFF
--- a/Documentation/SequentialDataStore/Searching.md
+++ b/Documentation/SequentialDataStore/Searching.md
@@ -1,3 +1,4 @@
+
 ---
 uid: sdsSearching
 ---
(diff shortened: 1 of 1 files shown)


Documentation/SequentialDataStore/Searching.md
```diff
@@ -72,6 +72,22 @@ After the previous call, you can use the following call to return the remaining
        _metadataService.GetStreamsAsync(“temperature*”, 100, 100)
```
mckenzielee: It looks like a table right under this is wrecked. Could you fix that as well?

ScottPatterson: should be fixed now

PR #10, opened 2018-11-14
Removing HTML tags and using <br> to wrap the values inside a single cell.

sukhdev1103:


ScottPatterson:
Looks good to me.  Now I know how to do multi-line rows in MarkDown tables :)

PR #26, opened 2019-01-03
Enable DocFX full-text search

hpaul-osi:
Adding the _enableSearch setting so that we can take advantage of the OOTB full-text search feature in DocFX.  The index will be automatically generated in the _site folder with each build.

### DIFF
--- a/docfx.json
+++ b/docfx.json
@@ -15,6 +15,9 @@
     }
   ],
   "build": {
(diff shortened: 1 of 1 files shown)


JeremyHenderson:
Bypassing Documentation Team as this is a functional change to the site to allow documentation search with no content change.

PR #36, opened 2019-01-17
Show the OSI icon in the header.

tabbott-osisoft:


JeremyHenderson:
Approving as it is a functional change not a content change.

PR #44, opened 2019-01-24
Updated with View to StreamView rename

mckenzielee:
Updated all reference of view to stream view including:
- SdsView ->SdsStreamView
- SdsViewProperty ->SdsStreamViewProperty
- SdsViewMode ->SdsStreamViewMode
- SdsViewMap ->SdsStreamViewMap
- SdsViewMapProperty ->SdsStreamViewMapProperty
- SdsTypeCodes have been modified to reflect the name changes above
- ISdsStreamReader has been modified to reflect the name changes above
- api/Tenants/{tenantId}/Namespaces/{namespaceId}/Views/{viewId}
-> api/Tenants/{tenantId}/Namespaces/{namespaceId}/StreamViews/{streamViewId}

- Task<SdsView>GetViewAsync(string viewId) ->
Task<SdsStreamView> GetStreamViewAsync(string streamViewId)


Documentation/SequentialDataStore/Reading_Data.md
```diff
@@ -91,16 +91,16 @@ Read characteristics applied to indexes between, before, or after data when the
...
+All reads support specifying a SdsStreamView identifier in the query string to shape the results of the read:
 
-``viewId={viewId}``
+``streamViewId={streamViewId}``
 
-Working with views is covered in detail in the [Sds Views](xref:sdsViews) section.
+Working with stream views is covered in detail in the [Stream Views](xref:sdsViews) section.
 
-When data is requested with a SdsView the read characteristics defined by the *target type* of the SdsView 
+When data is requested with a SdsView the read characteristics defined by the *target type* of the SdsStreamView
```
kanepenley: "When data is requested with an SdsStreamView" 
I think we need a comma after this part too?

kanepenley: I'm realizing we say "a SdsStream" and "a SdsType" everywhere rather than "an." I think "an" is right if you go by how it sounds, but maybe we should iron out which one we are supposed to use.

https://blog.apastyle.org/apastyle/2012/04/using-a-or-an-with-acronyms-and-abbreviations.html

Not sure if this is an actual authority on the issue, but I do think "an" is technically correct. We'd have to change it in a ton of places.

Documentation/SequentialDataStore/Reading_Data_API.md
```diff
@@ -1779,110 +1780,110 @@ Note that State is not included in the JSON as its value is the default value.
...
-   SdsBoundaryType boundaryType, string viewId = null);
+   SdsBoundaryType boundaryType, string streamViewId = null);
 
    Task<IEnumerable<T>> GetWindowFilteredValuesAsync<T>(string streamId, 
          string startIndex, string endIndex, SdsBoundaryType boundaryType, 
-         string filter, string viewId = null);
+         string filter, string streamViewId = null);
    Task<IEnumerable<T>> GetWindowFilteredValuesAsync<T, T1>(string streamId, 
-         T1 startIndex, T1 endIndex, SdsBoundaryType boundaryType, string filter, string viewId = null);
+         T1 startIndex, T1 endIndex, SdsBoundaryType boundaryType, string filter, string viestreamViewIdwId = null);
```
kanepenley: "viestreamViewIdwId" -> streamViewId

Documentation/SequentialDataStore/Reading_Data.md
mckenzielee: That looks right to me.

Documentation/SequentialDataStore/SDS_Views.md
```diff
@@ -658,19 +666,20 @@ If no matching identifier is found, the specified stream view is created.
...
+``string tenantId``  
+  The tenant identifier  
+  
+``string namespaceId``  
+  The namespace identifier  
+  
+``string streamViewId``  
+  The stream view identifier. The identifier must match the ``SdsStreamView.Id`` field.   
 
-The request content is the serialized SdsView. If you are not using the Sds client libraries, using JSON is recommended.
```
kanepenley: Shouldn't we be capitalizing SDS in places like the second sentence here?

Documentation/SequentialDataStore/SDS_Views.md
```diff
@@ -679,12 +688,12 @@ The request content is the serialized SdsView. If you are not using the Sds clie
...
 **Response body**
 
- The newly created or matching SdsView.
+ The newly created or matching SdsStreamView.
 
 
 **.NET Library**
 
-      Task<SdsView> GetOrCreateViewAsync(SdsView SdsView);
+      Task<SdsStreamView> GetOrCreateStreamViewAsync(SdsStreamView SdsStreamView);
```
kanepenley: if we want to match exactly how these methods are defined in ISdsMetadataService, technically the name of the argument should be sdsStreamView.

Documentation/SequentialDataStore/SDS_Views.md
mckenzielee: yep

Documentation/SequentialDataStore/SDS_Views.md
kanepenley: We do this in a lot of places though. Might be a separate work item to correct? (that might be the case for a couple of the suggestions I've made).

Documentation/SequentialDataStore/SDS_Views.md
```diff
@@ -200,16 +200,16 @@ automatically determine mapping.
     simple1Type.Name = "Simple1";
     simple1Type = await config.GetOrCreateTypeAsync(simple1Type);
 
-    SdsView view = new SdsView()
+    SdsStreamView view = new SdsStreamView()
```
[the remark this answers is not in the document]
mckenzielee: If we get complaints, I will change it.

PR #49, opened 2019-02-12
Omf ingress/remove publishers

mschellenger96:
Updating documentation to reflect change from publishers to clients for OMF ingress. Note that we should not check in this change until the code changes have been checked in.


Documentation/DataIngress/OMF_Ingress_Specification.md
```diff
@@ -12,10 +12,8 @@ OMF is interpreted by OSIsoft Cloud Services back-end system.
 Headers
 -------
 
-A description of each of the headers can be found in the [OMF spec](http://omf-docs.osisoft.com). When 
-sending messages to OSIsoft Cloud Services, the value of the ``producertoken`` header must be 
-set to a security token obtained from the OCS Portal. The security token is used to authenticate 
-the sender and to authorize the sender for use with a particular Tenant and Publisher.
+A description of each of the headers can be found in the [OMF spec](http://omf-docs.osisoft.com). Note that rather than using a ``producertoken``, data ingress calls to OCS require a bearer token to be attached in the header, as documented in the [OCS Quick Start](xref:sdsQuickStart) documentation. The bearer token is used to authenticate
```
osimarkrogers: "producertoken" or "producer token"? (like "bearer token")?

mschellenger96: "producertoken" is the header key value. Producer tokens are still part of the OMF spec, as they are used to send data to other back end systems, but they are no longer supported by OCS. The bearer token being referenced here is the same token that you would need to authenticate any other call to OCS. Publishers are being replaced with clients, but those client credentials are used to obtain a bearer token to attach to the headers to authenticate, if that makes sense

osimarkrogers: All good.  I will go ahead and merge then.

osimarkrogers:
Merged.

PR #51, opened 2019-02-20
Add data views doc

d-ng:


d-ng:
@osimarkrogers: fixed property descriptions according to comment.

PR #63, opened 2019-03-13
Fix singular model names in uris

bderveni:
- Fix singular model names in uris @osimarkrogers


osimarkrogers:
Reviewed, all consistent.

PR #72, opened 2019-04-02
Added information on nullable type interpolation.

mckenzielee:


Documentation/SequentialDataStore/SDS_Types.md
```diff
@@ -203,22 +203,24 @@ indexes that occur between data in a stream:
...
+|Boolean Type               |Returns value of nearest index  |         |
+|Enumeration Types          |Returns Enum value at 0         |This may have a value for the enumeration |
+|GUID                       |No event is returned            |         |
+|Version                    |No event is returned            |         |
+|IDictionary or IEnumerable |No event is returned            |Dictionary, Array, List, and so on. |
 
 *When extreme values are involved in an interpolation (for example
 Decimal.MaxValue) the call might result in a BadRequest exception.
 
+\**Nullable types are interpolated in the same manner as their non-nulllable equivalents as long as the events surrounding the desired interpolation index are non-null. If either of the indexes are null, the interpolated value will be null.
```
Robert-Wai: In "If either of the indexes are null", "indexes" is incorrect. It's the particular _value_ of the property of each event on either side of the interpolated index that's being interpolated, and the null check is on those values.

mckenzielee: you right.

PR #74, opened 2019-04-05
Remove page link in code block

antondahlin:


Documentation/DataViews/GroupRules.md
```diff
@@ -44,7 +44,7 @@ The following example shows a group rule definition that creates data groups bas
 ]
 ``` 
 ```json
-An example output using this group rule as described above could look like the following: (See [mappings](xref:Mappings) for a description on how to create the columns).
+An example output using this group rule as described above could look like the following:
```
osimarkrogers: New link needs to be provided, or this sentence introducing an example should be deleted.

antondahlin: The example being referred to is the output that follows the sentence, not the one in the link. That might have been unclear. The link was just to add more guidance on how to recreate the example, but it does not render in the Json tag, so it is easier to just not include that here.

osimarkrogers:
Understood.  Just merged.

Thanks Anton.

PR #91, opened 2019-05-07
Removed references to SdsType PUT APIs

Robert-Wai:


Documentation/SequentialDataStore/SDS_Types.md
```diff
@@ -33,9 +33,8 @@ characteristics when attempting to read non-existent indexes, indexes that fall
 existing indexes, are determined by the interpolation and extrapolation settings of the SdsType. For more 
 information about read characteristics see [Interpolation](#interpolation) and [Extrapolation](#extrapolation).
 
-SdsTypes are mostly immutable. When an SdsType is referenced by a stream or a stream view or another type, its form  cannot be changed. 
-Certain fields, such as the friendly name or description, can be changed because they do not affect the function 
-of the SdsType. In addition, the SdsType may be deleted only if no streams or stream views or types reference it.
+SdsTypes are immutable. After an SdsType is created, its definition cannot change. An SdsType must be deleted and recreated if the definition is incorrect.
+In addition, the SdsType may be deleted only if no streams or stream views or types reference it.
```
osimarkrogers: Should read "....deleted only if no streams, stream views, or types reference it."

Documentation/SequentialDataStore/SDS_Types.md
```diff
@@ -859,7 +858,7 @@ The new type may also include the full type definition of the reference type ins
 }
 ```
 
-If the full definition is sent, the referenced types (base type "Simple" in the above example) should match the actual type initially created, unless a type update is attempted. For type update behavior, see [Create Or Update Type](#create-or-update-type). If the full definition is sent and the referenced types did not exist, they will be created automatically by SDS. Further type creations can reference them as demostrated above. Note that when trying to get types back from SDS, the results will also include types that were automatically created by SDS.
+If the full definition is sent, the referenced types (base type "Simple" in the above example) should match the actual type initially created. If the full definition is sent and the referenced types did not exist, they will be created automatically by SDS. Further type creations can reference them as demonstrated above. Note that when trying to get types back from SDS, the results will also include types that were automatically created by SDS.
```
osimarkrogers: Change "did not exist" to "do not exist"

Documentation/SequentialDataStore/SDS_Types.md
Robert-Wai: fixed

Documentation/SequentialDataStore/SDS_Types.md
Robert-Wai: fixed

PR #98, opened 2019-05-30
Sampled Values

TomSeymour:
Note:  this includes the GetSampleFiltered methods which are coming soon.

### DIFF
--- a/Documentation/SequentialDataStore/Reading_Data.md
+++ b/Documentation/SequentialDataStore/Reading_Data.md
@@ -24,6 +24,7 @@ In addition, the following methods support reading multiple values:
 * [Get Interpolated Values](xref:sdsReadingDataApi#get-interpolated-values) retrieves a collection of stored or calculated values based on the request parameters.
 * [Get Summaries](xref:sdsReadingDataApi#get-summaries) retrieves a collection of evenly spaced summary intervals based on a count 
   and specified start and end indexes.
+* [Get Samples](xref:sdsReadingDataApi#get-samples) retrieves a collection of sampled data based on the request parameters.
 
 All single stream reads are HTTP GET actions. Reading data involves getting events from streams. The base reading URI from a single stream is as follows:
  ```text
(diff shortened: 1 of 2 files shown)


Documentation/SequentialDataStore/Reading_Data.md
```diff
@@ -146,6 +147,7 @@ Data transformations are supported for all single stream reads, but transformati
     api/v1-preview/Tenants/{tenantId}/Namespaces/{namespaceId}/Streams/{streamId}/Data/Transform
     api/v1-preview/Tenants/{tenantId}/Namespaces/{namespaceId}/Streams/{streamId}/Data/Transform/Interpolated
     api/v1-preview/Tenants/{tenantId}/Namespaces/{namespaceId}/Streams/{streamId}/Data/Transform/Summaries
+    api/v1-preview/Tenants/{tenantId}/Namespaces/{namespaceId}/Streams/{streamId}/Data/Transform/Sampled
```
mckenzielee: Do we support this? 
Would we apply the stream view, then the sample the transformed data? What happens if the sampleBy property is not in the target type?

mckenzielee:
@osimarkrogers We need to wait on the merge of this change.

Documentation/SequentialDataStore/Reading_Data_API.md
```diff
@@ -1267,6 +1267,126 @@ Content-Type: application/json
 ```
 ****
 
+## ``Get Samples``
+
+Returns data sampled by intervals between a specified start and end index. 
+  
+Sampling is driven by a specified Property or Properties of the stream's Sds Type.  Property types that cannot be interpolated do not support sampling requests. Strings are an example of indexes that cannot be interpolated.
```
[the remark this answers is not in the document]
TomSeymour: I believe the need to interpolate comes from the SdsBoundaryType (ExactOrCalculated)  which is a valid setting.  Additionally we are searching for maxs and mins, and in doing so we get the Double equivalent of the values before comparing them.  I am not sure what would happen with properties that couldn't be represented in a numerical form.

Documentation/SequentialDataStore/Reading_Data.md
gdinak: Yes.
Yes.
BadRequest.

PR #99, opened 2019-05-30
Update Topics + Subscriptions routes to v1-preview for GA

cchengosisoft:
We want to change Topics + Subscriptions routes temporarily from v1 to v1-preview as we go to GA. Once other teams are supporting v1, we can switch back to have v1 in our routes.


cchengosisoft:
abandoning PR as OCS has v1 available

PR #127, opened 2019-07-30
Doc updates for changes to GroupRule and Mappings objects

emonino:


Documentation/DataViews/Mappings.md
```diff
@@ -110,25 +96,25 @@ DataType | string | False | Type of the mapped data source
 ## `MappingRule` 
  The MappingRule can be used in the following forms 
  # [GroupRuleMappingRule](#tab/tabid-1) 
-The GroupMappingRule is used to map a column to a group rule by specifying the Id of the rule and a token defined in the group rule to identify what pattern to use.
+The GroupMappingRule is used to map a column to a group rule by specifying the Id of the rule and a group rule value defined in the group rule to identify what pattern to use.
```
osimarkrogers: comma after "Id of the rule"

Documentation/DataViews/Mappings.md
```diff
@@ -146,14 +132,9 @@ ItemIdentifier | FilterUnit | False | Identifier for the resource that contains
...
-```json
-{
-  "IsDefault": true
-}
-``` 
+The second example uses a shared mapping rule with a token to match the property path, and the item identifier including every stream with TankStream as its description. It also uses the group rule defined in the [Group Rules](xref:Mappings) page to create data groups based on the states of the data source. The FlowRate and Temperature columns are defined with the shared mapping rule, using the token to specify their property path. The Volume column is an example of explicitly defining the property mapping rule in the column definition. This is done to demonstrate the difference and the result would be identical using the shared rule and a token, as the previous columns do.
```
osimarkrogers: Delete comma between "path" and "and".  Add comma between "identifier" and "including"

osimarkrogers: Delete comma between "token" and "as".  Add comma between "identical" and "using".

Documentation/DataViews/Mappings.md
emonino: Fixed punctuation

Documentation/DataViews/Mappings.md
emonino: Fixed punctuation

emonino: Fixed punctuation

PR #146, opened 2019-09-03
Summary support for nullable types

sukhdev1103:


Documentation/SequentialDataStore/Reading_Data_API.md
```diff
@@ -1241,6 +1241,103 @@ Content-Type: application/json
 ]
 ```
 
+## ``Get Summaries for Nullable Types``
+
+OCS also support summary requests for nullable SdsTypes. It means a SdsType has at least a nullable SdsTypeProperty.
```
jchan-osi: For the first sentence, it should be "OCS also support[s]". I also think we should use SDS instead of OCS as the subject.
For the second sentence. It should be "a[n] SdsType" instead of "a SdsType".

Documentation/SequentialDataStore/Reading_Data_API.md
```diff
@@ -1241,6 +1241,103 @@ Content-Type: application/json
...
+      11/23/2017 12:00:02 PM: Measurement 2
+      11/23/2017 12:00:03 PM: Measurement null
+      11/23/2017 12:00:04 PM: Measurement 1
+      11/23/2017 12:00:05 PM: Measurement 2
+	  11/23/2017 12:00:06 PM: Measurement null
+	  11/23/2017 12:00:07 PM: Measurement null
+	  11/23/2017 12:00:08 PM: Measurement 3
+			
+
+The following request calculates one summary intervals between the `startIndex` and `endIndex`:
```
jchan-osi: "one summary interval" instead of "one summary intervals".

Documentation/SequentialDataStore/Reading_Data_API.md
```diff
@@ -1241,6 +1241,103 @@ Content-Type: application/json
 ]
 ```
 
+## ``Get Summaries for Nullable Types``
```
mckenzielee: This should not be a separate section.

Documentation/SequentialDataStore/Reading_Data_API.md
```diff
@@ -1285,7 +1285,7 @@ Content-Type: application/json
 SDS also supports summary requests for nullable SdsTypes. It means an SdsType has at least a nullable SdsTypeProperty.
 
 **Example** 
-The following example contains a nullable double property with interpolation mode as continous.
+The following example contains a nullable double property with interpolation mode as continuous.
```
GiselleLeeOSI: Maybe instead of 'as', use 'set to'?
Also use colon (:) in place of period (.) since you're introducing an example.

Documentation/SequentialDataStore/Reading_Data_API.md
```diff
@@ -1577,6 +1578,21 @@ Index identifying the beginning of the series of events to return
 ``string endIndex``  
 Index identifying the end of the series of events to return
 
+``int count``  
+Optional maximum number of events to return.
+
+``SdsBoundaryType boundaryType``  
+Optional SdsBoundaryType specifies the handling of events at or near the startIndex and endIndex
```
GiselleLeeOSI: Question: are the three boundary types mutually exclusive of each other? If boundaryType is specified, for example, do you automatically not need startBoundaryType or endBoundaryType?

Documentation/SequentialDataStore/Reading_Data_API.md
```diff
@@ -1262,6 +1280,130 @@ Content-Type: application/json
...
+        "Start": {
+            "Time": "2017-11-23T12:00:01Z",
+            "Measurement": 2
+        },
+        "End": {
+            "Time": "2017-11-23T12:00:08Z",
+            "Measurement": 3
+        },
+        "Summaries": {
+            "Count": {
```
jchan-osi: What is the difference between Time and Measurement? Is one the number of events and the other is number of intervals? I am not sure do we need to explain them or not.

Documentation/SequentialDataStore/Reading_Data_API.md
```diff
@@ -1262,6 +1280,130 @@ Content-Type: application/json
...
+	  11/23/2017 12:00:06 PM: Measurement null
+	  11/23/2017 12:00:07 PM: Measurement null
+	  11/23/2017 12:00:08 PM: Measurement 3
+
+**Note:** Non-weighted summaries disregard null values and treat them as non-existent. In above example, non-weighted summaries for Measurement would be calculated based on (2,2,1,2,3). 
+
+Weighted summaries consider null values for its calculation. 
+While calculating weighted summaries, if we encounter a null value at a given index then we would consider interpolation mode of property to find the interpolated value of the given interval. Please see the below table for [12:00:02 PM, 12:00:03 PM] interval. The values are 2 at 12:00:02 PM and null at 12:00:03 PM. 
+
+| Interpolation Mode | Weight in Seconds | Value in meter |
```
jchan-osi: This "Seconds" starts with an upper case "s" while the others start with a lower case "s".
Does meter need to be plural? I am not sure.

Documentation/SequentialDataStore/Reading_Data_API.md
sukhdev1103: Time is index property and Measurement is field property. So Count would tell you how many times Time property has appeared and how many times Measurement property appeared in an interval.

public class SimpleType
{
   [SdsMember(IsKey = true, Order = 0) ]
   public DateTime **Time** { get; set; }
   [SdsMember(Uom = "meter")]
   public double? **Measurement** { get; set; }
}

Documentation/SequentialDataStore/Reading_Data_API.md
sukhdev1103: i was also not sure. I will either change both to plural or to singular.

Documentation/SequentialDataStore/Reading_Data_API.md
jchan-osi: Ok, got it. So Measurement is 4 instead of 5 because 4 intervals are measured?

sukhdev1103: yes, right. But Kane is working on fixing the count. We should show it as 5 becuase there are actual 5 events. Internally, we know there are 4 intervals so summaries call return back 4 as count.

jchan-osi: So we should probably use the fixed version for the docs.

sukhdev1103: yes, I will change it.

Documentation/SequentialDataStore/Reading_Data_API.md
```diff
@@ -1262,6 +1280,131 @@ Content-Type: application/json
...
+| ContinuousNullableLeading <br> StepwiseContinuousLeading | 0 | 0 | 
+| ContinuousNullableTrailing <br> StepwiseContinuousTrailing | 1 | 1 |
+
+| Interpolation Mode | Weight in seconds | Value in meter |
+| ---------- | ----------------------- | ------------- |
+| Continuous | 1 | 1.5 |
+| ContinuousNullableLeading <br> StepwiseContinuousLeading | 1 | 1 | 
+| ContinuousNullableTrailing <br> StepwiseContinuousTrailing | 1 | 2 |
+
+For more information see [Interpolation.](xref:sdsTypes#interpolation)
```
GiselleLeeOSI: @sukhdev1103 Is there a difference between having the period inside the bracket instead of outside? If not, I think punctuation should not be part of the link.

sukhdev1103: Yes you are right. there should not be the period

GiselleLeeOSI:
Doc looks good. Please merge @mckenzielee.

PR #147, opened 2019-09-04
Update OcsVariableSupport.dll to support 'long URL wrap-breaking by words'

WenjieLi-osi:
Update OcsVariableSupport.dll to support 'long URL wrap-breaking by words'


JeremyHenderson:
Funtional improvement on OCS-Docs. Approving for that reason.

PR #152, opened 2019-09-25
Update OcsVariableSupport.dll

potrusil-osi:
Use <wbr> for line break opportunities.


JeremyHenderson:
Merged as this is a functional change and not a content change to DocFX

PR #156, opened 2019-10-02
Fixed summary count and explained calculations

kanepenley:
I added explanations on which summary calculations are event weighted vs index weighted, as well as fixing the summary counts and adding the new index count to the summary.


kanepenley:
This change needs to sync with a push to production, so it shouldn't be merged yet.

PR #158, opened 2019-10-15
Update documentation for Access Rights routes

dleeatosisoft:


Documentation/DataIngress/OMF_Ingress_Subscriptions.md
```diff
@@ -93,6 +93,24 @@ Get the default Access Control List for new Subscriptions.
 
 An AccessControlList object.
 
+*****************
+
+``GET api/v1/tenants/{tenantId}/namespaces/{namespaceId}/accessrights/subscriptions``
+--------------------------------------------
+
+Get the default Access Rights of the requesting identity for new Subscriptions.
```
mschellenger: I think this could be re-worded a bit. I get what you are trying to say but I had to read it a few times for it to make sense. Maybe something like "Get the default Access Rights of the requesting identity for any newly created Subscription"

Documentation/DataIngress/OMF_Ingress_Topics.md
```diff
@@ -108,6 +108,24 @@ An AccessControlList object.
 
 ***************************
 
+``GET api/v1/tenants/{tenantId}/namespaces/{namespaceId}/accessrights/topics``
+-------------------------------------------------------------------------------------
+
+Get the default Access Rights of the requesting identity for any newly created Topics.
```
mschellenger: Nitpicky, but you use plural "Topics" here and use singular "Subscription" in the equivalent route. I think either are fine

Documentation/DataIngress/OMF_Ingress_Subscriptions.md
dleeatosisoft: Changed.

Documentation/DataIngress/OMF_Ingress_Topics.md
dleeatosisoft: I changed it to use the plural.

PR #169, opened 2019-12-02
Updating SDS section with type edits

GiselleLeeOSI:
Spelling out the first instance of the portal as "OSIsoft Cloud Service portal".

### DIFF
--- a/Documentation/SequentialDataStore/Searching.md
+++ b/Documentation/SequentialDataStore/Searching.md
@@ -4,8 +4,7 @@ uid: sdsSearching
 
 Searching
 =====================
-Search in SDS provides a way to search text, fields, etc. across the Sequential Data Store. This document covers the 
-searching for SdsStreams, SdsTypes, and SdsStreamViews.
+Using the REST API and client libraries method, you can search for fields in streams, types and stream views across SDS.
 
 Searching for Streams
 =====================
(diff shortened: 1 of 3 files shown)


Documentation/SequentialDataStore/Searching.md
```diff
@@ -4,8 +4,7 @@ uid: sdsSearching
 
 Searching
 =====================
-Search in SDS provides a way to search text, fields, etc. across the Sequential Data Store. This document covers the 
-searching for SdsStreams, SdsTypes, and SdsStreamViews.
+Using the REST API and client libraries method, you can search for fields in streams, types and stream views across SDS.
```
mckenzielee: Since the rest of the sentence goes on the describe capabilities of different rest calls and multiple client libraries methods. I suggest changing to "Using the REST APIs and client libraries,"

Also, does "Using the REST API **and** client libraries method" imply the user needs to use both? Users will typically pick one or the other. 

This seems to imply that you can only search for fields in streams, types, and stream views. That is not the case.  The user specifies the search query and we return all streams, types, and stream views that match the search query parameters. You can search for streams, types, and stream views by phrases and fields.

GiselleLeeOSI:
Not sure how to remove certain files from a branch. Closing this PR and will start the work off fresh.

PR #183, opened 2020-02-14
Update Data_View_API.md

pragnapasunuri:
We have changed the behavior of our PUT endpoint to align with that of SDS's behavior. This documentation change involves details about the new behavior


d-ng:
Looks good from the technical perspective. Note to others: we will only merge this when we are ready to push the related changes to production.

PR #187, opened 2020-02-26
Updating resource Id requirement on spacing for types, streams, and stream views

GiselleLeeOSI:
Updating resource Id requirement (WI 92099), Combining with WI 97403 (minor copy edits to types and streams topics)


Documentation/SequentialDataStore/SDS_Streams.md
```diff
@@ -31,9 +31,8 @@ for internal SDS use.
 they are associated with SdsStream objects and can be used as search criteria.
 
 **Rules for the Stream Identifier (SdsStream.Id)**
-
 1. Is not case sensitive
-2. Can contain spaces
+2. Is not blank and cannot contain leading or trailing whitespace
```
FreeJules: I would separate this into two different cases
- Is not blank (not sure if blank is the right choice of words... can be interpreted as nothing at all vs. whitespaces only, what do you think?)
- cannot contain leading or trailing whitespace

Documentation/SequentialDataStore/SDS_Types.md
```diff
@@ -5,43 +5,42 @@ uid: sdsTypes
...
-events. Events are stored in streams, called SdsStreams. An SdsType defines the shape or structure of the 
-event and how to associate events within the SdsStream.
+events. Events are stored in SdsStreams (or streams). An SdsType (or type) defines the shape or structure of the 
+event and how to associate events within the stream.
 
 
-SdsTypes can define simple atomic types, such as integers, floats, strings, arrays, and dictionaries, or 
-they can define complex types using SdsTypes. You can define complex, nested types using the Properties 
-collection of an SdsType. 
+You can define simple atomic types, such as integers, floats, strings, arrays, and dictionaries, or
```
FreeJules: Are we allowed to use 'you' in documentation, if we are this is fine, otherwise I'd rephrase it to 'Types can be defined as simple etc.'

Documentation/SequentialDataStore/SDS_Types.md
```diff
@@ -5,43 +5,42 @@ uid: sdsTypes
...
-SdsTypes define how events are associated and read within a collection of events, or SdsStream. The read 
+SdsTypes define how events are associated and read within a collection of events (SdsStream, that is). The read 
 characteristics when attempting to read non-existent indexes, indexes that fall between, before or after 
 existing indexes, are determined by the interpolation and extrapolation settings of the SdsType. For more 
-information about read characteristics see [Interpolation](xref:sdsReadingData#interpolation) and [Extrapolation](xref:sdsReadingData#extrapolation).
+information about read characteristics, see [Interpolation](xref:sdsReadingData#interpolation) and [Extrapolation](xref:sdsReadingData#extrapolation).
 
-SdsTypes are immutable. After an SdsType is created, its definition cannot change. An SdsType must be deleted and recreated if the definition is incorrect.
-In addition, the SdsType may be deleted only if no streams, stream views, or types reference it.
+SdsTypes are immutable. After a type is created, its definition cannot change. A type must be deleted and recreated if the definition is incorrect.
```
FreeJules: 'cannot be changed'? kinda sounds better to me but I think it's ok like this too. You decide.

Documentation/SequentialDataStore/SDS_Types.md
```diff
@@ -51,39 +50,35 @@ The following table shows the required and optional SdsType fields. Fields that
...
-| Properties        | IList\<SdsTypeProperty\> | Required    | Yes, with limitations | List of SdsTypeProperty items |
-
-
-**Rules for the Type Identifier (SdsType.Id)**
+| Properties        | IList\<SdsTypeProperty\> | Required    | Yes, with limitations | List of SdsTypeProperty items. See [SdsTypeProperty](#sdstypeproperty) below.  |
 
+### Rules for the Type Identifier (SdsType.Id)
 1. Is not case sensitive
-2. Can contain spaces
+2. Is not blank and cannot contain leading or trailing whitespace
```
FreeJules: Same as above... I'd separate this...

Documentation/SequentialDataStore/SDS_Types.md
```diff
@@ -597,7 +585,7 @@ var simpleType = new SdsObjects.SdsType({
 });
 ```
 
-Working with a derived class is easy. For the following derived class:
+You can also work with a derived class easily. This is a derived class:
```
FreeJules: 'You' again... Maybe rephrase this without it... or not... :)

Documentation/SequentialDataStore/SDS_Types.md
```diff
@@ -609,7 +597,7 @@ class Derived(Simple):
         self.__observation = observation
 ```
 
-Extend the SdsType as follows:
+You can then extend the above SdsType by doing this:
```
FreeJules: here too

Documentation/SequentialDataStore/SDS_Views.md
```diff
@@ -38,7 +38,7 @@ The following table shows the required and optional SdsStreamView fields. Fields
 **Rules for the Stream View Identifier (SdsStreamView.Id)**
 
 1. Is not case sensitive
-2. Can contain spaces
+2. Is not blank and cannot contain leading or trailing whitespace
```
FreeJules: same here as above two

Documentation/SequentialDataStore/SDS_Types.md
GiselleLeeOSI: I ran it by the writers just to be sure but yes, using an active voice is preferred. And "you" is okay to use in docs.

Documentation/SequentialDataStore/SDS_Types.md
GiselleLeeOSI: Yes it's okay to use but it's not absolutely necessary here. See if you like the fix better.

PR #191, opened 2020-03-03
Renaming

sturnberg:
Renaming section to group and distinguisher to identifying field


d-ng:
To @osimarkrogers : please don't merge this after review. This needs to be coordinated with our push to production with the changes. Thanks!

sturnberg:
I created a local branch and made the changes to all of the pages and then ran:
>git push --set-upstream origin sturnberg-data-view-rename
Unfortunately, I didn't have permission to create a branch.  Since that didn't work I edited the pages individually using the documentation system.  A single branch that would maintain our changes until we're ready to go to production would be ideal but I'm not sure we can do that.

PR #207, opened 2020-04-02
Add links to the customer portal in support page

GiselleLeeOSI:
Adding links per discussion with Javin


GiselleLeeOSI:
Updated support page, changed title in the index and toc.yml

PR #227, opened 2020-05-26
updated field mapping count limit

BethZelasky:
Update documentation with short explanation of data field mapping count limit.

### DIFF
--- a/Documentation/DataViews/Getting_Data.md
+++ b/Documentation/DataViews/Getting_Data.md
@@ -111,4 +111,4 @@ By default, requests for a first page of data will cause the data view to re-res
 Requests for subsequent pages include a `continuationToken`. This implictly suppresses re-resolution, akin to cache "preserve" behavior.
 
 ## Field mapping count limit
-Data views are limited to 1000 field mappings to ensure timely access of data. The field mapping check is made prior to constructing the data set.
+Data views are limited to 1000 field mappings to ensure timely access of data. OCS checks the field mapping count prior to constructing the data set.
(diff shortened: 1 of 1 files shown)


Documentation/DataViews/Getting_Data.md
```diff
@@ -109,3 +109,6 @@ It is recommended to suppress [re-resolution](xref:ResolvedDataView#when-is-a-da
 By default, requests for a first page of data will cause the data view to re-resolve. See the documentation on [resolved data views](xref:ResolvedDataView#when-is-a-data-view-resolved). This ensures that the data view accounts for any streams that have been added to or removed from SDS. Re-resolution may be suppressed by explicitly specifying a cache behavior of "preserve".
 
 Requests for subsequent pages include a `continuationToken`. This implictly suppresses re-resolution, akin to cache "preserve" behavior.
+
+## Field mapping count limit
```
osimarkrogers, line 113: This should probably be "OCS checks the field mapping before constructing the data set." (If that is a technically correct statement)

Documentation/DataViews/Getting_Data.md
```diff
@@ -109,3 +109,6 @@ It is recommended to suppress [re-resolution](xref:ResolvedDataView#when-is-a-da
...
+Data views are limited to 1000 field mappings to ensure timely access of data. The field mapping check is made prior to constructing the data set.
```
osimarkrogers, line 114: This should probably be "OCS checks field mapping prior to constructing the data set." OR "You should check field mapping prior to constructing the data set." (whichever is correct)

PR #231, opened 2020-06-03
Standardizing format

mckenzielee:
Adopted the tags and metadata docs to the new standardized format.


GiselleLeeOSI:
I took a look and made a few more fixes where applicable. Changes include:
1) Updated header Sample response body to Example response body to match the proposed template
2) Relabeled the security header to authorization per template and placed it before response (per template)

PR #244, opened 2020-07-14
Properties table template

GiselleLeeOSI:
Creating a PR to have the draft for properties table committed.


vtouati1009:
1) I lean toward using field name because "property name" seems vague.
2) Description seems more accurate.

Documentation/Properties_Table_Template.md
```diff
@@ -0,0 +1,13 @@
+---
+uid: ocsPropertiesTable
+---
+|*Column name*| Field/ ``Property Name`` | Data Type | Required? | Default | Searchable | Details/ ``Description`` |
+|-------------|-------|------|----------|---------|------------|---------|
+|*What it is*|    name of the property   | data type of the property     |   Is this property optional or required?       |  default value or format       |   Is the property searchable?         |   description of the property      |
+|*Optional?*|       |      |   optional column       |   optional column      |      optional column      |         |
+|*Example*|  Id     | String     | Yes         |         |     Yes       |    Unique identifier     |
+
```
lasato-OSI, line 9: I find this table confusing. I think it needs some work in terms of presentation of the information. 
Although I see that you want to consolidate the description in the table itself, I think this is confusing. I would create a table and include a couple of examples in it. Keep the description of the table columns separate.

Documentation/Properties_Table_Template.md
```diff
@@ -0,0 +1,13 @@
...
+Questions for writers:
+1. Do you prefer calling it a `field name` or `property name`?
```
lasato-OSI, line 11: I think "field" is what we call the text fields in the user interface. I think "property name" is more typically used, but I defer to your judgment.

Documentation/Properties_Table_Template.md
```diff
@@ -0,0 +1,13 @@
...
+2. What about `details` vs `description`?
```
lasato-OSI, line 12: Description

Documentation/Properties_Table_Template.md
GiselleLeeOSI: This is a very good point. I put it together the way it is now so I could talk about at the meeting but yes, I'll redo it after I get all your feedback. Description and examples in a separate location, that is.

Documentation/Properties_Table_Template.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Example properties table
+| Field/ ``Property Name`` | Data Type | Required | Default | Searchable | Details/ ``Description`` |
+|-------|------|----------|---------|------------|---------|
+|  Id     | String     | Yes         |         |     Yes       |    Unique identifier     |
+|  GroupingFields     | Field[]     | No        |  []       |     Yes       |   Fields by which the data items are partitioned/grouped     |
+
+## Questions for writers
+Questions for writers:
+1. Do you prefer calling it a `field name` or `property name`?
```
osimarkrogers, line 22: I vote for "property name" since it's in a "properties table". Monotony is a good thing in term usage.

Documentation/Properties_Table_Template.md
```diff
@@ -0,0 +1,24 @@
...
+2. What about `details` vs `description`?
```
osimarkrogers, line 23: Description

GiselleLeeOSI:
Questions for writers:
1. Do you prefer calling it a `field name` or `property name`?  Field:1 Property: 2 Either: 1
2. What about `details` vs `description`? Desc: all

GiselleLeeOSI:
@mckenzielee  and @d-ng, I created a properties table template and was hoping you'd review it and give feedback. As you can see from the conversation above, it was reviewed by writers already.

Documentation/Properties_Table_Template.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Table description
+| Property Name | Data Type | Default Type | Required | Searchable | Description |
+|-------|------|----------|---------|------------|---------|
+|    name of the property   | data type of the property     |   default value or format      |   this property optional or required?       |   Is the property searchable?         |   description of the property      |
+|       |      |   This is an optional column       |   This is an optional column      |      This is an optional column      |         |
+
+
+## Example properties table
+| Property Name | Data Type |  Default Type | Required | Searchable | Description |
```
mckenzielee, line 15: Instead of Default Type it should be Default Value, meaning if you do not specify a value for this property this is the value we will set it to. Data Type and Default Type are confusing because they seem to mean the same thing. I think you could also use another example that more clearly demonstrates this, like an object with an enumeration property.

If the property is required then the default value column will be empty because there is no default. I would consider swapping those two columns for that fact.

GiselleLeeOSI: Thank you, updated the column name  and swapped the two column orders. Can you point me to the example which includes enumeration property?

Documentation/API-template.md
```diff
@@ -0,0 +1,85 @@
...
+``` 
+
+### Parameters
+`datatype resource1Id`   
+resource1Id description. `resourceId` should be camelCase. 
+
+[Optional] `datatype resource2Id`   
+[Optional] resource2Id description. No need to tag the parameter if it is required. 
+
+If there is more than one type of parameter, specify each one. Use H3 header (three hashtags). See examples below.
```
d-ng, line 24: Clarification question - the `Request path parameters` and `Request query parameters` will replace the `Parameters` section, right?

Documentation/API-template.md
```diff
@@ -0,0 +1,85 @@
...
+```text
+Examples
+### Request path parameter
```
d-ng, line 27: should this plural "parameters"?

Documentation/API-template.md
```diff
@@ -0,0 +1,85 @@
...
+### Request query parameters
+```
+
+#### Request body 
+Description of request body
```
d-ng, line 32: This section should also be optional. (e.g. GET request)

Documentation/API-template.md
```diff
@@ -0,0 +1,85 @@
...
+Code example can follow. 
+
+| Status Code | Response Type | Description | 
+|--|--|--| 
+| 200 OK | `Object` | Object description | 
+| 403 Forbidden | error | You are not authorized for this operation | 
+| 404 Not Found | error | The data view or query does not exist | 
+| 500 Internal Server Error | error | An error occurred while processing the request. | 
+
+```json
```
d-ng, line 59: What sort of code example goes here? Initially, I thought it would contain the example response body, but there is a section on example response body below.

Documentation/API-template.md
```diff
@@ -0,0 +1,85 @@
...
+#### Response headers 
+ This section is optional. Add any important information about headers. 
+
+Successful (200 OK) responses include one or more header values related to paging. 
+| Header | Description | 
+|--|--| 
+| Header1 | Description of Header1 | 
+| Header2 | Description of Header2 | 
+
+#### Response body
```
d-ng, line 73: This section should also be optional (if no response body)

Documentation/API-template.md
```diff
@@ -0,0 +1,85 @@
...
+#### Example response body
+This section is optional.
+```json 
+Put code example here. 
+``` 
+
+*** Add a line before each API. 
+To see APIs formatted according to the template, see [Get Stream (SdsStreams API > Get Stream)](https://ocs-docs.osisoft.com/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.html#get-stream)
+or [Get Data Views (Data view API > Get Data Views)](https://ocs-docs.osisoft.com/Content_Portal/Documentation/DataViews/DataViewsAPIOverview/Data_View_API.html#get-data-views).
+**Note:** Links might break while OCS-docs is going through an update.
```
d-ng, line 85: Any conclusion on the formatting of the .NET client libraries method? Are we removing them from the API references?

Documentation/Properties_Table_Template.md
```diff
@@ -0,0 +1,21 @@
...
+|-------|------|----------|---------|------------|---------|
+|    Name of the property   | Data type of the property     |    Is this property optional or required?      | Default value if the property value is not set        |   Is the property searchable?         |   Description of the property      |
+|       |      |   This is an optional column       |   This is an optional column      |      This is an optional column      |         |
+
+
+## Example properties table
+| Property Name | Data Type |  Required | Default Value | Searchable | Description |
+|-------|------|----------|---------|------------|---------|
+|  Id     | String     |     Yes     |        |     Yes       |    Unique identifier     |
+|  Tag     | String[]     |         |   Array       |            |    Gets or sets for OSIsoft internal use only     |
```
d-ng, line 18: nitpicking here :P but the default value of "Array" is a little misleading. If they are empty arrays the data views doc use `[]`.

Documentation/API-template.md
GiselleLeeOSI: Unfortunately, not yet. I'll reach out to Konstantin and let you know if I hear back.

Documentation/Properties_Table_Template.md
GiselleLeeOSI: This is a good point. Thank you.

GiselleLeeOSI:
Closing without merge. Feedback received from necessary parties.

PR #246, opened 2020-07-20
Change description of the "count" query param in GetData

mktfos:
We now calculate default page size instead of hard-coding it as 1,000.

The PR reflects this changes in the description of the query parameter "count" for both preview and non-preview GetData endpoints.

### DIFF
--- a/Content_Portal/Documentation/DataViews/Data_API.md
+++ b/Content_Portal/Documentation/DataViews/Data_API.md
@@ -47,7 +47,7 @@ Controls when the data view backing resources are to be refreshed. Used only whe
 | `Preserve`| Use cached information, if available.   
 
 `[optional] int count`  
-The requested page size. The default value is 1000. The maximum is 250,000.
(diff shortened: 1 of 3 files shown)


mktfos:
@GiselleLeeOSI that's a good point. I changed the description, although I'm still not 100% sure it is very understandable. And ok, I'll add Mark.

mktfos:
@GiselleLeeOSI 
@osimarkrogers 
Hi Mark, I was trying to add you as a reviewer here but for some reason your name doesn't show up in the list of reviewers in the sidebar. Can you please take a look at the doc change? Thanks.

d-ng:
@mktfos: Probably worth mentioning somewhere that the count is returned as the header on the next page/first page in case users want to find out the page size. I also wonder if it's useful to add additional/detailed information on the getting data page directly, and keep the description on the API page minimal.

mktfos:
@d-ng that's a good point. I added this information to the data page and only kept the calculation info on the API pages. Or should I move the calculation details to the data page too?

PR #248, opened 2020-07-21
Adding a line on how compound indexing of the primary index affects secondary indexes

GiselleLeeOSI:
Adding a line as part of the WI 136155.

If the primary index is a compound index, secondary indexes will only be valid if the compound index has a maximum of 2 properties.

### DIFF
--- a/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
+++ b/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
@@ -48,7 +48,7 @@ Secondary indexes are defined on the SdsStream while the primary index is define
 If the primary index defined on the SdsType is a compound index, secondary indexes on the SdsStream are allowed as long as that compound index does not have more than 2 properties. For more information on compound indexes, see [Indexes](xref:sdsIndexes#compound-indexes). 
 <!-- Secondary indexes apply to a single property. In other words, there are no compound secondary indexes.-->
   
-Note that you can only use the SdsTypeCodes that can be ordered (DateTime or numbers, for example) as a secondary index.
+Note that you can only use the SdsTypeCodes of SdsTypes whose property can be ordered (DateTime or numbers, for example) as a secondary index.
 
 ## Interpolation and extrapolation
 The InterpolationMode, ExtrapolationMode, and [PropertyOverrides](#propertyoverrides) can be used to determine how a specific SdsStream reads data.
--- a/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
+++ b/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
@@ -44,7 +44,7 @@ However, they are associated with SdsStream objects and can be used as search cr
 5. Can contain a maximum of 100 characters
 
 ## Indexes
-You define secondary indexes on the SdsStream and primary index on the SdsType.
+You define secondary indexes on the SdsStream and the primary index on the SdsType.
 If the primary index defined on the SdsType is a compound index, secondary indexes on the SdsStream are allowed as long as that compound index does not have more than two properties. For more information on compound indexes, see [Indexes](xref:sdsIndexes#compound-indexes). 
 <!-- Secondary indexes apply to a single property. In other words, there are no compound secondary indexes.-->
   
--- a/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
+++ b/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
@@ -45,7 +45,8 @@ However, they are associated with SdsStream objects and can be used as search cr
 
 ## Indexes
 While the primary index is defined at the SdsType, secondary
-indexes are defined at the SdsStream.
+indexes are defined at the SdsStream. If the primary index is a compound index,
+secondary indexes will only be valid if the compound index has a maximum of 2 properties.
 
 Secondary indexes are applied to a single property; there are no
 compound secondary indexes. Only SdsTypeCodes
--- a/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
+++ b/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
@@ -48,7 +48,7 @@ You define secondary indexes on the SdsStream and the primary index on the SdsTy
 If the primary index defined on the SdsType is a compound index, secondary indexes on the SdsStream are allowed as long as that compound index does not have more than two properties. For more information on compound indexes, see [Indexes](xref:sdsIndexes#compound-indexes). 
 <!-- Secondary indexes apply to a single property. In other words, there are no compound secondary indexes.-->
   
-Note that you can only use the SdsTypeCodes of SdsType properties that can be ordered (DateTime or numbers, for example) as a secondary index.
+Note that you can only use the SdsTypeCodes of SdsType properties that can be ordered (``DateTime`` or numbers, for example) as a secondary index.
 
 ## Interpolation and extrapolation
 The InterpolationMode, ExtrapolationMode, and [PropertyOverrides](#propertyoverrides) can be used to determine how a specific SdsStream reads data.
--- a/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
+++ b/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
@@ -45,8 +45,7 @@ However, they are associated with SdsStream objects and can be used as search cr
 
 ## Indexes
 Secondary indexes are defined on the SdsStream while the primary index is defined on the SdsType.
-Secondary indexes apply to a single property. In other words, there are no compound secondary indexes.
-Secondary index is allowed as long as the primary index that is a compound index has up to 2 properties.   
+If the primary index defined on the SdsType is a compound index, secondary indexes on the SdsStream are allowed as long as that compound index does not have more than 2 properties. Secondary indexes apply to a single property. In other words, there are no compound secondary indexes.
   
 Note that only the SdsTypeCodes that can be ordered are supported for use in a secondary index.
 For more information, see [Indexes](xref:sdsIndexes).
--- a/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
+++ b/Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
@@ -44,7 +44,7 @@ However, they are associated with SdsStream objects and can be used as search cr
 5. Can contain a maximum of 100 characters
 
 ## Indexes
-You define secondary indexes on the SdsStream and the primary index on the SdsType.
+While you define the primary index on the SdsType, the SdsStream is where you define secondary indexes.
 If the primary index defined on the SdsType is a compound index, secondary indexes on the SdsStream are allowed as long as that compound index does not have more than two properties. For more information on compound indexes, see [Indexes](xref:sdsIndexes#compound-indexes). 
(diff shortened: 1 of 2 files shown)


GiselleLeeOSI:
secondary indexes are only valid for compound indexes with up to two properties

GiselleLeeOSI:
Original text: 
While the primary index is defined at the SdsType, secondary indexes are defined at the SdsStream.

Secondary indexes are applied to a single property; there are no compound secondary indexes. Only SdsTypeCodes that can be ordered are supported for use in a secondary index.

For more information on indexes, see Indexes.

GiselleLeeOSI:
@annavillena, take a look and let me know what you think. I pretty much rewrote the whole Indexes section. I hope the original meaning was not altered in the process. See above for the original text.  As discussed, please mention this at the standup and find an appropriate second set of eyes. Thank you for reviewing.

Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
```diff
@@ -44,14 +44,12 @@ However, they are associated with SdsStream objects and can be used as search cr
...
-indexes are defined at the SdsStream.
-
-Secondary indexes are applied to a single property; there are no
-compound secondary indexes. Only SdsTypeCodes
-that can be ordered are supported for use in a secondary index.
-
-For more information on indexes, see [Indexes](xref:sdsIndexes).
+Secondary indexes are defined on the SdsStream while the primary index is defined on the SdsType.
+Secondary indexes apply to a single property. In other words, there are no compound secondary indexes.
+Secondary index is allowed as long as the primary index that is a compound index has up to 2 properties.
```
annavillena, line 49: I think this is confusing because it makes it seem like secondary indexes are only allowed in this case. I suggest: If the primary index defined on the SdsType is a compound index, secondary indexes on the SdsStream are allowed as long as that compound index does not have more than 2 properties.
I would also make this the second sentence so it connects to the definition of "primary index"

GiselleLeeOSI: I agree. Bumped your suggested sentence to the second position.

Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
```diff
@@ -44,14 +44,11 @@ However, they are associated with SdsStream objects and can be used as search cr
...
-While the primary index is defined at the SdsType, secondary
-indexes are defined at the SdsStream.
-
-Secondary indexes are applied to a single property; there are no
-compound secondary indexes. Only SdsTypeCodes
-that can be ordered are supported for use in a secondary index.
-
-For more information on indexes, see [Indexes](xref:sdsIndexes).
+Secondary indexes are defined on the SdsStream while the primary index is defined on the SdsType.
+If the primary index defined on the SdsType is a compound index, secondary indexes on the SdsStream are allowed as long as that compound index does not have more than 2 properties. Secondary indexes apply to a single property. In other words, there are no compound secondary indexes.
```
FreeJules, line 48: I would remove this sentence. It seems a bit confusing to me... I had to read it couple of times to process it. I would maybe break it down and add to compound index section in indexes that 'compound index only applies to type index' and in secondary indexes that 'even though indexes on a stream are defined as a list, each element of the list can be only a single property, so you can't specify multiple properties to single secondary index'. Or something like that. What do you think?

annavillena: I agree that the last sentence is misleading. The Compound Index section already states that compound indexes are only for indexes defined at SdsType level. My suggestion for the sentence here: "There can be multiple secondary indexes per stream defined as a list of indexes. Each element of the list is an individual property and will not be converted to a compound index. Compound indexes are only applicable for primary indexes defined on the SdsType."

Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
```diff
@@ -44,14 +44,11 @@ However, they are associated with SdsStream objects and can be used as search cr
...
-Secondary indexes are applied to a single property; there are no
-compound secondary indexes. Only SdsTypeCodes
-that can be ordered are supported for use in a secondary index.
-
-For more information on indexes, see [Indexes](xref:sdsIndexes).
+Secondary indexes are defined on the SdsStream while the primary index is defined on the SdsType.
+If the primary index defined on the SdsType is a compound index, secondary indexes on the SdsStream are allowed as long as that compound index does not have more than 2 properties. For more information on compound indexes, see [Indexes](xref:sdsIndexes#compound-indexes). 
+<!-- Secondary indexes apply to a single property. In other words, there are no compound secondary indexes.-->
+  
+Note that you can only use the SdsTypeCodes that can be ordered (DateTime or numbers, for example) as a secondary index.
```
annavillena, line 51: I think I know what this SdsTypeCodes note is missing. It should be "Note that you can only use properties with SdsTypeCodes that can be ordered (DateTime or numbers, for example) as secondary indexes."

dwicinas-osisoft:
@GiselleLeeOSI I spent a little time looking at this PRPR (pull request peer review) and it quickly became apparent that I have _no idea_ how to handle it. I think it's premature to ask me to do a peer review of any real documentation in the form of a pull request. Perhaps I'm a slow student and I should understand the process better after three whole training sessions. But, sadly, looking at this pull request, I have no clue. That's the sad reality of the situation.

Adding @osifweber

Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
```diff
@@ -44,14 +44,11 @@ However, they are associated with SdsStream objects and can be used as search cr
...
 ## Indexes
-While the primary index is defined at the SdsType, secondary
-indexes are defined at the SdsStream.
-
-Secondary indexes are applied to a single property; there are no
-compound secondary indexes. Only SdsTypeCodes
-that can be ordered are supported for use in a secondary index.
-
-For more information on indexes, see [Indexes](xref:sdsIndexes).
+Secondary indexes are defined on the SdsStream while the primary index is defined on the SdsType.
```
dwicinas-osisoft, line 47: Could line 47 be rephrased to avoid passive voice. I'm not sure, because I don't quite understand how you create indexes. Is it the system creating the index, in which case passive voice would be acceptable. Or, is it the user, in which case you should avoid passive voice, if possible.

Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
```diff
@@ -44,14 +44,11 @@ However, they are associated with SdsStream objects and can be used as search cr
...
+If the primary index defined on the SdsType is a compound index, secondary indexes on the SdsStream are allowed as long as that compound index does not have more than 2 properties. For more information on compound indexes, see [Indexes](xref:sdsIndexes#compound-indexes).
```
dwicinas-osisoft, line 48: Write out numbers less than 10. Thus, "2" should be "two."

Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
```diff
@@ -44,14 +44,11 @@ However, they are associated with SdsStream objects and can be used as search cr
...
-Secondary indexes are applied to a single property; there are no
-compound secondary indexes. Only SdsTypeCodes
-that can be ordered are supported for use in a secondary index.
-
-For more information on indexes, see [Indexes](xref:sdsIndexes).
+Secondary indexes are defined on the SdsStream while the primary index is defined on the SdsType.
+If the primary index defined on the SdsType is a compound index, secondary indexes on the SdsStream are allowed as long as that compound index does not have more than 2 properties. For more information on compound indexes, see [Indexes](xref:sdsIndexes#compound-indexes). 
+<!-- Secondary indexes apply to a single property. In other words, there are no compound secondary indexes.-->
+  
+Note that you can only use the SdsTypeCodes of SdsTypes whose property can be ordered (DateTime or numbers, for example) as a secondary index.
```
dwicinas-osisoft, line 51: This may be more of a style thing, but I would avoid using a personal pronoun like "whose" when referring to something like a property. The IBM Style Guide says it's acceptable, but it gives the reader pause. Also, can you avoid the passive in that sentence?

Maybe:

Note that you can only use the SdsTypesCodes of SdsType with a property that can be ordered ..... 

or 

Note that you can only use the SdsTypesCodes of SdsType with a property that you can order .... (to avoid passive voice)

Content_Portal/Documentation/SequentialDataStore/indexes.md
```diff
@@ -63,7 +63,10 @@ In read and write operations, specify compound indexes in the URI by ordering ea
  separated by the pipe character, ‘|’. 
 To help those using compound indexes, .NET client libraries methods also allow the use of tuples for indexes.
 
-**Notes:** The examples below are for compound indexes on SdsTypes and not of secondary indexes on SdsStreams.
+**Notes:** 
+- Compound indexing only applies to SdsTypes. In other words, there is no compound indexing for secondary indexes which are on SdsStreams. For more information see [SdsStreams](xref:sdsStreams#indexes).
```
dwicinas-osisoft, line 67: small issue:

secondary indexes **that** are on SdsStreams. See IBM Style Guide, page 29 for a quick explanation about relative pronouns. Better yet, read E. B. White's _Elements of Style_.

(Have I mentioned that I used to teach English composition?)

Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
GiselleLeeOSI: I've updated in the latest commit

Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
GiselleLeeOSI: Went with the first option.

Content_Portal/Documentation/SequentialDataStore/indexes.md
GiselleLeeOSI: Updated. Also added a comma after "For more information"

Content_Portal/Documentation/SequentialDataStore/SDS_Streams.md
```diff
@@ -44,14 +44,11 @@ However, they are associated with SdsStream objects and can be used as search cr
...
 ## Indexes
-While the primary index is defined at the SdsType, secondary
-indexes are defined at the SdsStream.
-
-Secondary indexes are applied to a single property; there are no
-compound secondary indexes. Only SdsTypeCodes
-that can be ordered are supported for use in a secondary index.
-
-For more information on indexes, see [Indexes](xref:sdsIndexes).
+You define secondary indexes on the SdsStream and the primary index on the SdsType.
```
annavillena, line 47: I feel like the original sentence 'While the primary index...' made the contrast between secondary and primary indexes clearer.

GiselleLeeOSI: We (writers) are told  to avoid using passives. Who defines indexes? Is it done by the person or the system? If it is indeed done by the system, then I will revert it to passive again.

annavillena: I see. You're correct, it's done by the person. Could it be: While you define a primary index at the SdsType, you can define secondary indexes at the SdsStream.
I started using the word 'at' too because that is how it is said in the Indexes section.

annavillena: If you think it makes more sense with the current change, though, I'm okay with that.

PR #249, opened 2020-07-22
Adding two rows to the interpolation table

GiselleLeeOSI:
Adding two rows to the interpolation table for object type and empty type


GiselleLeeOSI:
@FreeJules, as discussed, I'll just keep the table simple for now.

PR #251, opened 2020-08-14
Lasato ocs docs reorg

lasato-OSI:
I cleaned up my files to eliminate extraneous warning messages.


Content_Portal/toc.yml
```diff
@@ -1,5 +1,7 @@
 - name: OSIsoft Cloud Services
```
osimarkrogers, line 1: The OSIsoft Cloud Services and Quickstarts links are both wrong.

A toc link to an md file should be a topicHref, and a toc link to a nested toc should be an href.
So the OSIsoft Cloud Services link should be:

- name: OSIsoft Cloud Services
- topicHref: OCS_Content_Portal_Overview.md

The Quickstarts link needs a landing page, because the toc.yml file it's pointing to will only display in the left hand menu, and on the right you would have a blank page.  So this should look something like:

- name: Quickstarts
- href: Quickstarts/toc.yml
- topicHref: Quickstarts Overview.md (imaginary file - - would also need its location if not in same folder as this level toc, like Quickstarts/Quickstarts Overview.md)

The API Reference link is fine as is...good example to follow.

lasato-OSI: I am aware of these anomalies. There is no relevant content to include at the moment.

PR #264, opened 2020-09-18
HTTP status codes

lasato-OSI:
Hi Ji Won,

Please review and approve this pull request.

Thanks,
Laureen


GiselleLeeOSI:
Laureen, everything looks good. Checking out other tables in the current documentation, wrapping might be handled already. Just one thing: I see that you changed the file name to use dash instead of underscore, which is good. But it still doesn't match the title in the topic (H1: HTTP status codes). This came up during yesterday's url/domain meeting so as a good practice we probably should conform with the new rule of using lowercase and dashes for the file name. So it should be: http-status-codes.

PR #277, opened 2020-09-30
Digitaljoshua patch getsetproperties

digitaljoshua:
Removing Get and Set from properties descriptions
Adding/updating all properties table to template spec


digitaljoshua:
Starting over from master - no need to add extra optional columns in properties tables

PR #286, opened 2020-10-06
Update Available_Field_Sets.md

osimarkrogers:
Added Note and Link to Retrieve Available Field Sets to show origin of retrieved field sets in example.

### DIFF
--- a/Content_Portal/Documentation/DataViews/DefineaDataView/Available_Field_Sets.md
+++ b/Content_Portal/Documentation/DataViews/DefineaDataView/Available_Field_Sets.md
@@ -28,7 +28,10 @@ The example uses the following data view and the streams from the [Stream exampl
 }
 ```
 
-Accessing the ```Get Available Field Sets``` endpoint will show available fields corresponding to stream id, name, tags, metadata, and each property. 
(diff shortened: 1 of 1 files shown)


Content_Portal/Documentation/DataViews/DefineaDataView/Available_Field_Sets.md
```diff
@@ -28,7 +28,10 @@ The example uses the following data view and the streams from the [Stream exampl
 }
 ```
 
-Accessing the ```Get Available Field Sets``` endpoint will show available fields corresponding to stream id, name, tags, metadata, and each property. 
+Accessing the ```Get Available Field Sets``` endpoint will show available fields corresponding to stream id, name, tags, metadata, and each property.
+
+**Note:** The field sets returned are the field sets included in the example PUT command in the Recommended Workflow.[Recommended Workflow](xref:DefineaDataViewRecommendedWorkflow#include-some-of-the-available-fields)
```
d-ng, line 33: This is slightly confusing because updating the data field set should come after retrieving available field sets. Taking a step back, what is the purpose of this addition? I might be better equipped to give a suitable suggestion that way.

osimarkrogers: My apologies - - I misread the text where I inserted the link.  Closing this PR.

PR #287, opened 2020-10-07
Proposal for Streamlining Recommended Workflow

osimarkrogers:
Linking from Recommended Workflow to actual procedure to eliminate duplication.

### DIFF
--- a/Content_Portal/Documentation/DataViews/DefineaDataView/Available_Field_Sets.md
+++ b/Content_Portal/Documentation/DataViews/DefineaDataView/Available_Field_Sets.md
@@ -30,6 +30,14 @@ The example uses the following data view and the streams from the [Stream exampl
 
 Accessing the ```Get Available Field Sets``` endpoint will show available fields corresponding to stream id, name, tags, metadata, and each property. 
 
+### Action
+```text
+GET /api/v1/Tenants/{tenantId}/Namespaces/{namespaceId}/DataViews/quickstart/Resolved/AvailableFieldSets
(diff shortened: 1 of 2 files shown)


Content_Portal/Documentation/DataViews/DefineaDataView/Available_Field_Sets.md
```diff
@@ -30,6 +30,14 @@ The example uses the following data view and the streams from the [Stream exampl
 
 Accessing the ```Get Available Field Sets``` endpoint will show available fields corresponding to stream id, name, tags, metadata, and each property. 
 
+### Action
+```text
+GET /api/v1/Tenants/{tenantId}/Namespaces/{namespaceId}/DataViews/quickstart/Resolved/AvailableFieldSets
```
d-ng, line 35: This Uri (and response) will not work if user is not going through the steps in the recommended workflow as they are listed. I am hesitant to accept the changes as is. The recommended workflow is written as a quick start, that user need to follow from start to end. If we were to simplify it and spread out the content to different section, we should comprehensively design and rewrite the content.

osimarkrogers: Right - - I'd like to smooth that out.  I do think that the Recommended Workflow would be much better if all it presented was the actual workflow, just the sequence of operations, with just a brief description of the operation/step, and a link to each full procedure.  And we can normalize all the procedure's examples so they work.

PR #298, opened 2020-10-26
updated code font

tmartin-osisoft:
Hi, I made a minor change to the font that governs the color for code. 
An example of this is in the Asset API topic (Laureen, don't worry, I'm only pushing this one CSS file and not anything related to Assets).  Check out the Get Asset ID H2 heading in both production and CLGUILD to see the difference. 

We used to use a reddish font color and that is currently being used for OCS in production. I've changed the color to the dark grey that we've adopted for Edge/Adapter docs.


rmuolic (approved):

GiselleLeeOSI (approved):
Looks good

PR #307, opened 2020-11-16
Update Consent.md

digitaljoshua:
Divya, I am updating the consent topic with information about two consent email in the consent process. Please let me know if I did this correctly. I am unsure about the the verbiage in the second consent for advanced integration, as I don't have an idea what it looks like, and if there is another Consent button.

### DIFF
--- a/Content_Portal/Concepts/Consent.md
+++ b/Content_Portal/Concepts/Consent.md
@@ -5,17 +5,16 @@ uid: identityConsent
 # AAD Consent to OSIsoft Applications
 
 You have chosen to use your organization’s existing Azure Active Directory (AAD) to sign in to your
-OSIsoft Cloud Services (OCS) account. In order to proceed, you first need to complete a consent workflow. In
-this step, your AAD administrator will grant permissions for OCS Identity to do the following:
+OSIsoft Cloud Services (OCS) account. In order to proceed, you must complete the consent process to allow your OCS tenant to access your AAD tenant and directory information for authentication. This consent process involves your AAD administrator granting permissions for OCS Identity to sign in and read user profile basic information (delegated permission).
 
-- Sign in and read user profile basic information (delegated permission).
+Completing the neccesary consent process involves these steps:
 
-Below are the next steps required to complete your OCS account provisioning:
-
-- An email will be sent to your AAD administrator who has the privileges to grant OCS access to your AAD.
-- Your administrator will click on the link provided in the email. Note that your administrator will need to log in as a user with the Global Administrator role on your AAD.
-- Upon login, your administrator will be prompted to grant consent for a single application: OCS Identity.
-- Your administrator should select the box "Consent on behalf of the organization" before clicking the consent button.
-- OSIsoft will get a confirmation that the consent process has been completed.
-- Then, a second email will be sent to the initial user who signed up for the account.
-- At that point, this initial user will be able to use the link provided to activate his profile and sign in your OCS account.
+1. For a new tenant, an email will be sent to your AAD administrator who has the privileges to grant OCS access to your AAD. For an existing tenant with access to your AAD tenant, skip to step 6. Note that if your administrator agrees with only one of the two consent emails required, your account will remain in a partially-consented state.
+2. Your AAD administrator will click on the link provided in the email which will bring them to a login page. Your administrator will need to log in as a user with the Global Administrator role on your AAD.
+3. Upon successful login, your administrator will be prompted to grant consent for a single application: OCS Identity.
+4. Your administrator should select the box "Consent on behalf of the organization" before clicking the consent button.
+5. OSIsoft will get a confirmation that the consent process has been completed.
+6. Your administrator will get an email about consent to advanced integration with your OCS tenant and your AAD information. 
+7. Your administrator should agree to the advanced integration by clicking the consent button.
+8. A final email will be sent to the initial user who signed up for the account, informing them that they can sign in with their AAD account.
+7. At that point, this initial user will be able to use the link provided to activate his profile and sign in your OCS account.


Content_Portal/Concepts/Consent.md
```diff
@@ -5,17 +5,16 @@ uid: identityConsent
 # AAD Consent to OSIsoft Applications
 
 You have chosen to use your organization’s existing Azure Active Directory (AAD) to sign in to your
-OSIsoft Cloud Services (OCS) account. In order to proceed, you first need to complete a consent workflow. In
-this step, your AAD administrator will grant permissions for OCS Identity to do the following:
+OSIsoft Cloud Services (OCS) account. In order to proceed, you must complete the consent process to allow your OCS tenant to access your AAD tenant and directory information for authentication. This consent process involves your AAD administrator granting permissions for OCS Identity to sign in and read user profile basic information (delegated permission).
```
gavchen, line 8: If Advanced Integration is enabled for the AAD identity provider, then the following permissions would also need to be consented:
•	Read all users' full profiles
•	Read all group memberships

Content_Portal/Concepts/Consent.md
```diff
@@ -5,17 +5,16 @@ uid: identityConsent
...
+2. Your AAD administrator will click on the link provided in the email which will bring them to a login page. Your administrator will need to log in as a user with the Global Administrator role on your AAD.
+3. Upon successful login, your administrator will be prompted to grant consent for a single application: OCS Identity.
+4. Your administrator should select the box "Consent on behalf of the organization" before clicking the consent button.
+5. OSIsoft will get a confirmation that the consent process has been completed.
+6. Your administrator will get an email about consent to advanced integration with your OCS tenant and your AAD information. 
+7. Your administrator should agree to the advanced integration by clicking the consent button.
+8. A final email will be sent to the initial user who signed up for the account, informing them that they can sign in with their AAD account.
+7. At that point, this initial user will be able to use the link provided to activate his profile and sign in your OCS account.
```
gavchen, line 20: Step number may need fixing.

Content_Portal/Concepts/Consent.md
```diff
@@ -5,17 +5,16 @@ uid: identityConsent
...
-
-- An email will be sent to your AAD administrator who has the privileges to grant OCS access to your AAD.
-- Your administrator will click on the link provided in the email. Note that your administrator will need to log in as a user with the Global Administrator role on your AAD.
-- Upon login, your administrator will be prompted to grant consent for a single application: OCS Identity.
-- Your administrator should select the box "Consent on behalf of the organization" before clicking the consent button.
-- OSIsoft will get a confirmation that the consent process has been completed.
-- Then, a second email will be sent to the initial user who signed up for the account.
-- At that point, this initial user will be able to use the link provided to activate his profile and sign in your OCS account.
+1. For a new tenant, an email will be sent to your AAD administrator who has the privileges to grant OCS access to your AAD. For an existing tenant with access to your AAD tenant, skip to step 6. Note that if your administrator agrees with only one of the two consent emails required, your account will remain in a partially-consented state.
```
gavchen, line 12: For my understanding, OCS 'tenant' is synonymous with OCS 'account', right?

Should we expand upon what a partially-consented state account means? For a newly created tenant that had AAD configured as the identity provider and Advanced Integration enabled, no users will be able to login to the account in a partially-consented state. For an existing tenant which already had AAD configured as an identity provider but just enabled Advanced Integration and is waiting for the consent for its required permissions, pre-existing users can still access the account while it is in the partially-consented state. New users can also still be added for the AAD identity provider, but without the features of Advanced Integration (by search).

Content_Portal/Concepts/Consent.md
gavchen, line 20: The way the steps are written now, they are assuming the OCS tenant is provisioned at creation with both AAD as an identity provider and Advanced Integration enabled. However, we should maybe clarify the steps to reflect the three possible paths from the starting point of a new tenant: 

1. Choosing to provision a new tenant with an identity provider which is not AAD (AAD + Advanced Integration can be added separately later).
2. Choosing to provision a new tenant with AAD as an identity provider without Advanced Integration (Advanced Integration can be added later).
3. Choosing to provision a new tenant with AAD and Advanced Integration.

For case 1, any possible future consent needed would separately be for [sign in, read profile] (AAD) and [read users' full profiles, read all group memberships] (Advanced Integration). If AAD and Advanced Integration are added separately and not at the same time, then two separate consent emails will be sent, each with one link. If they are added at the same time, then just one email with two consent links will be sent out.

For case 2, a consent email with one link will be sent when the new tenant is provisioned with AAD as an identity provider. If Advanced Integration is later added, then another email will be sent with one more link.

For case 3, one email will be sent with two consent links.

Content_Portal/Concepts/Consent.md
```diff
@@ -5,17 +5,16 @@ uid: identityConsent
...
+2. Your AAD administrator will click on the link provided in the email which will bring them to a login page. Your administrator will need to log in as a user with the Global Administrator role on your AAD.
+3. Upon successful login, your administrator will be prompted to grant consent for a single application: OCS Identity.
+4. Your administrator should select the box "Consent on behalf of the organization" before clicking the consent button.
```
gavchen, line 15: I'm not sure if this has changed or if I may not be looking at the right thing, but I don't see a box to select, nor the verbiage "Consent on behalf of the organization". What I see is the following, after clicking the consent link in the email and signing in:

image

Content_Portal/Concepts/Consent.md
```diff
@@ -5,17 +5,16 @@ uid: identityConsent
...
+5. OSIsoft will get a confirmation that the consent process has been completed.
+6. Your administrator will get an email about consent to advanced integration with your OCS tenant and your AAD information. 
+7. Your administrator should agree to the advanced integration by clicking the consent button.
```
gavchen, line 18: As stated in the comment below for lines 12 to 20, there should just be one email with two consent links in the case that both AAD is added as an identity provider and Advanced Integration is enabled at the time of tenant creation (case 3). Cases 1 and 2 would have separate consent emails sent, but the rest of the steps should be clarified for these other 2 possible cases and flows.

I'm not sure what the agreed-upon casing is for "Advanced Integration", but I have seen it capitalized in the UI and since it is a feature name, perhaps it should be capitalized?

gavchen (changes requested):
Thanks for writing this up! I have posted my comments - let me know if there are any questions. I also only recently learned what this flow looks like visually, so let me know if you want to go through any of it together.

digitaljoshua:
This PR is a branch of another branch that involves restructuring of API pages which no longer will happen. therefore I created a new PR branch #309

PR #323, opened 2020-12-17
Update toc.yml

lasato-OSI:
Changed Assets to Asset Store in toc.


lasato-OSI:
Updating terminology in toc.

PR #343, opened 2021-01-27
Fixing inaccurate response description on 404

GiselleLeeOSI:
Fixing inaccurate response description on 404 throughout SDS API documentation. Specifically, it should be resource, not data view or query. Missed while reviewing ACL Patch API addition.


FreeJules (approved):
``

PR #394, opened 2021-03-18
Incorporating Product Readiness feedback into roles, users, clients topics

lasato-OSI:


Content_Portal/Overview/get-started/gs-roles.md
```diff
@@ -17,8 +17,18 @@ Follow the procedure to create role.
 
 1. In the toolbar, click **Add Role**.
 
-1. Enter a Name for the role and, optionally, a Description.
+1. In the Add Role window, enter the following:
+   - Name  &ndash; MyOCS.Role
+   - Description &ndash; Optionally, add a description for this role
+   
+1. Click **Add** to create the new role. It will be added to the table of Roles. Note that the Role Type is listed as *Account Custom* <!-- verify that this is still Account -->for all custom (or non-default) roles. You can now specify permissions for this role when managing permissions on assets, resources, and services in OCS.
```
dcbroad3, line 24: It looks like all of the Role Types are still listed as "Account". The Role *Names* have been updated to tenant, but the Role *Types* still use the old Account terminology.
Might be good feedback to the devs, though, as I'm sure that's not intended.

Content_Portal/Overview/get-started/gs-users.md
```diff
@@ -16,12 +16,14 @@ Follow the procedure to invite a user to your tenant.
 
 1. If multiple identity providers are enabled in your tenant, you are prompted to select an identity provider. Choose the appropriate identity provider for the user you are adding. After you select an option, you can later switch to a different option.
 
-1. Enter a **First Name** and **Last Name** for the user.  
+1. Enter a first and last name  for the user.
```
dcbroad3, line 19: I think there's an extra space here after `last name`

lasato-OSI, line 19:
```suggestion
1. Enter a first and last name for the user.  
```

lasato-OSI: Done

Content_Portal/Overview/get-started/gs-roles.md
lasato-OSI: John Sintilas verified that role types should appear as "Tenant <role_name>.

dcbroad3: @lasato-OSI does that mean that they have a Work Item to update those names? They are definitely still "Account" today, but definitely makes sense to use the new names in the documentation if the update is imminent.
image

lasato-OSI: We got confirmation from the Soteria team that this change is in Integration and hasn't made it into the Production server yet. So I expect the Production server will be updated to Tenant before April 20th which is when this goes live. But I'll be sure to verify this in the portal and follow up if necessary.

PR #400, opened 2021-03-22
Added Josh's feedback about ACLs and roles

lasato-OSI:
Hi Doug and Derek,

One of the OCS writers raised a question about how we talk about roles being the mechanism by which users' permissions to OCS resources is determined. The question is whether we should be clearer that it is a combination of roles and ACLs that determine whether a user has access to a resource or not.

Please take a look at Josh's comments and see what you think. Thanks

### DIFF
--- a/Content_Portal/set-up/roles-concept.md
+++ b/Content_Portal/set-up/roles-concept.md
@@ -5,7 +5,6 @@ uid: ccRoles
 
 Roles are used to manage access to assets, resources, and services in OSIsoft Cloud Services (OCS). Roles are assigned to identities, which includes users, groups, and client-credentials clients. When an identity attempts to access a resource in OCS, the identity's list of roles is compared against the permissions on the resource to determine whether access is allowed.
 
-<!--- Josh Kim Mar19021: Similar to the comment as before where it seemed to me that this sentence implied roles were the primary factor in getting access, when in fact it is the roles of the users compared against the provisioned ACL that grants access. Do we want to communicate this? --->
 
 There are five built-in roles which cannot be removed from a tenant.
 
--- a/Content_Portal/set-up/roles-concept.md
+++ b/Content_Portal/set-up/roles-concept.md
@@ -3,7 +3,9 @@ uid: ccRoles
 ---
 # About roles
 
-Roles are used to manage access to assets, resources, and services in OSIsoft Cloud Services (OCS). Roles are assigned to identities, which includes users, groups, and client-credentials clients. <!--- Josh Kim Mar19021: Similar to the comment as before where it seemed to me that this sentence implied roles were the primary factor in getting access, when in fact it is the roles of the users compared against the provisioned ACL that grants access. Do we want to communicate this? --->
+Roles are used to manage access to assets, resources, and services in OSIsoft Cloud Services (OCS). Roles are assigned to identities, which includes users, groups, and client-credentials clients. 
+
+<!--- Josh Kim Mar19021: Similar to the comment as before where it seemed to me that this sentence implied roles were the primary factor in getting access, when in fact it is the roles of the users compared against the provisioned ACL that grants access. Do we want to communicate this? --->
 
 There are five built-in roles which cannot be removed from a tenant.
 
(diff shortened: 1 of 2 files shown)


dcbroad3:
I think it's a fair point. We don't want to give the impression that to manage security you only have to manage a user's roles. It might be implied that the roles have hard-coded access, which is not the case. Josh's clarification that users get access by comparing roles against the ACL is valuable. I don't know if we want to use the term "ACL" in a document like this, might say something simpler like "Roles are assigned to identities. When an identity attempts to interact with a resource in OCS, the identity's list of roles is compared against the permissions on the resource to determine whether access is allowed."

Content_Portal/set-up/roles-concept.md
```diff
@@ -3,7 +3,9 @@ uid: ccRoles
 ---
 # About roles
 
-Roles are used to manage access to assets, resources, and services in OSIsoft Cloud Services (OCS). Roles are assigned to identities, which includes users, groups, and client-credentials clients. <!--- Josh Kim Mar19021: Similar to the comment as before where it seemed to me that this sentence implied roles were the primary factor in getting access, when in fact it is the roles of the users compared against the provisioned ACL that grants access. Do we want to communicate this? --->
+Roles are used to manage access to assets, resources, and services in OSIsoft Cloud Services (OCS). Roles are assigned to identities, which includes users, groups, and client-credentials clients.
```
lasato-OSI, line 6:
```suggestion
Roles are used to manage access to assets, resources, and services in OSIsoft Cloud Services (OCS). Roles are assigned to identities, which includes users, groups, and client-credentials clients. When an identity attempts to access a resource in OCS, the identity's list of roles is compared against the permissions on the resource to determine whether access is allowed.
```

Content_Portal/set-up/roles-concept.md
```diff
@@ -3,7 +3,9 @@ uid: ccRoles
 ---
 # About roles
 
-Roles are used to manage access to assets, resources, and services in OSIsoft Cloud Services (OCS). Roles are assigned to identities, which includes users, groups, and client-credentials clients. <!--- Josh Kim Mar19021: Similar to the comment as before where it seemed to me that this sentence implied roles were the primary factor in getting access, when in fact it is the roles of the users compared against the provisioned ACL that grants access. Do we want to communicate this? --->
+Roles are used to manage access to assets, resources, and services in OSIsoft Cloud Services (OCS). Roles are assigned to identities, which includes users, groups, and client-credentials clients. 
+
+<!--- Josh Kim Mar19021: Similar to the comment as before where it seemed to me that this sentence implied roles were the primary factor in getting access, when in fact it is the roles of the users compared against the provisioned ACL that grants access. Do we want to communicate this? --->
```
lasato-OSI, line 8:
```suggestion

```

Content_Portal/set-up/roles-procedure.md
```diff
@@ -5,6 +5,10 @@ uid: gpRoles
 
 Roles are used to manage access to assets, resources, and services in OSIsoft Cloud Services (OCS). Roles are assigned to identities, which includes users, groups, and client-credentials clients. See the following for more information about roles:
```
lasato-OSI, line 6:
```suggestion
Roles are used to manage access to assets, resources, and services in OSIsoft Cloud Services (OCS).  See the following for more information about roles:
```
Josh: What do you think if I make this introduction really brief and point them to the  "About roles" topic for more information. I'm trying to avoid repeating the same content in multiple places.

Content_Portal/set-up/roles-procedure.md
```diff
@@ -5,6 +5,10 @@ uid: gpRoles
...
 
+ <!-- Josh Kim: Roles are assigned to identities. It is the OCS object, specifically the ACL, that determines whether the identity gets access. This is done by comparing the assigned role of the identity to the OCS object's ACL. I wonder if the first sentence gives the impression that Roles is only factor in gaining access? -->
```
lasato-OSI, line 8:
```suggestion
```

Content_Portal/set-up/roles-procedure.md
```diff
@@ -5,6 +5,10 @@ uid: gpRoles
...
+
+or perhaps this should be our recommended message to customers to simplify access? Could we ask readiness/architects how detailed we want to be about describing access?
```
lasato-OSI, line 10:
```suggestion

```

Content_Portal/set-up/roles-procedure.md
dcbroad3: You have an extra space after the period in your suggestion. Otherwise I'm OK with this.

PR #405, opened 2021-03-23
Configure stream views

GiselleLeeOSI:
1. Added uid to what does ocs do topic to link to it.
2. Created a task topic for configuring stream views.

### DIFF
--- a/Content_Portal/add-organize-data/configure-stream-views.md
+++ b/Content_Portal/add-organize-data/configure-stream-views.md
@@ -4,8 +4,13 @@ uid: task-configureStreamViews
 
 # Configure stream views
 Stream views provide flexibility in the use of types and allow you to expose data as a different type. 
-While you cannot actually change the properties of types themselves, the stream views feature enables you to create a view of a selected stream that appears as if you had changed the type on which it is based. You create a stream view by choosing a source and target type then a set of mappings between properties of those two types.
-For more information, see [stream views](xref:WhatOCSdoes#stream-views). To configure stream views, follow this procedure:
+While you cannot actually change the properties of types themselves, the stream views feature enables you to create a view of a selected stream that appears as if you had changed the type on which it is based.
+You can use stream views to change the appearance of the data stream while keeping the archived events intact. 
+You create a stream view by choosing a source and target type and mapping properties of the two types.
+Note that you can only map properties of the same data type (double with int16, for example).
+
+For more information, see [stream views](xref:WhatOCSdoes#stream-views).
+If you already have types and streams configured with data, go directly to step 4:
 
 1. Create types. Go to **Data Management** > **Sequential Data Store**.
   - Click **Types**.
@@ -23,28 +28,27 @@ For more information, see [stream views](xref:WhatOCSdoes#stream-views). To conf
       - Add property.
         - Add *timestamp-target*, check **Key Id** and select *DateTime* as **Type**.
         - Add *measurement*, select *Int16* as **Type**. 
-2. Create streams.
+
+2. Create a stream based on the source type.
   - Click **Streams**.
   - Create a stream based on the source type.
     - Click **+ Add Stream**. 
-    - Add *source-stream* as **Stream Id**.
-    - Select *source-type* from **Type Id**.
-  - Create another stream based on the target type. 
-    - Click **+ Add Stream**.
-    - Add *target-stream* as **Stream Id**.
-    - Select *target-type* from **Type Id**
-3. Add data into the streams. 
+    - In **Stream Id**, add *source-stream*.
+    - From **Type Id**, select *source-type*.
+ 
+3. Add data into the stream. 
   - Check the box next to *source-stream*.
     - Click **Data**.
     - Add data into *source-stream*.
       - Click **+ Add Event**.
       - Add data. 
-  - Click **Streams** again. 
-  - Check the box next to *target-stream*.  
-    - Click **Data**.  
-    - Add data into *target-stream*.
-      - Click **+ Add Event**.
-      - Add data. 
+   ```json
+ { "timestamp": "2021-01-01T00:00:00Z", "value": 113.1 },
+ { "timestamp": "2021-01-18T02:00:00Z", "value": 15  }, 
+ { "timestamp": "2021-02-3T04:00:00Z", "value": -24.9, }, 
+ { "timestamp": "2021-02-208T06:00:00Z", "value": 0.4 }, 
+ { "timestamp": "2021-03-8T08:00:00Z", "value": 2001.2}
+  ```
 4. Create a stream view.
    - Click **Stream Views**.
    - Click **+ Add Stream View**.
@@ -56,10 +60,21 @@ For more information, see [stream views](xref:WhatOCSdoes#stream-views). To conf
      - Click **+ Property Mapping**.
      - On the left of the arrow, select **value** from the list.  
      - On the right of the arrow, select **measurement** from the list.
+   
+        **Note:** Mapped properties must be of the same data type. 
    - Click **Save**.
 5. Check how stream view transformed your data in the API console. 
    - Go to **Developer Tools** > **API Console**.
    - In **URI**, enter the following endpoint in the URI and click **Get**.
 ```text
-/Namespaces/{yourNamespace}/Streams/source-stream/Data/Transform?startIndex={earliestDateWithData}&count=100&streamViewId=stream-view
+/Namespaces/{yourNamespace}/Streams/source-stream/Data/Transform?startIndex=2021-01-01&count=100&streamViewId=stream-view
 ```
+  - Note changes in the displayed data.
+  
+```json
+ { "timestamp-target": "2021-01-01T00:00:00Z", "measurement": 113 },
+ { "timestamp-target": "2021-01-18T02:00:00Z", "measurement": 15  }, 
+ { "timestamp-target": "2021-02-3T04:00:00Z", "measurement": -24, }, 
+ { "timestamp-target": "2021-02-208T06:00:00Z", "measurement": 0 }, 
+ { "timestamp-target": "2021-03-8T08:00:00Z", "measurement": 2001}
+``` 
--- a/Content_Portal/add-organize-data/configure-stream-views.md
+++ b/Content_Portal/add-organize-data/configure-stream-views.md
@@ -1,0 +1,65 @@
+---
+uid: task-configureStreamViews
+---
+
+# Configure stream views
+Stream views provide flexibility in the use of types and allow you to expose data as a different type. 
+While you cannot actually change the properties of types themselves, the stream views feature enables you to create a view of a selected stream that appears as if you had changed the type on which it is based. You create a stream view by choosing a source and target type then a set of mappings between properties of those two types.
+For more information, see [stream views](xref:WhatOCSdoes#stream-views). To configure stream views, follow this procedure:
+
+1. Create types. Go to **Data Management** > **Sequential Data Store**.
+  - Click **Types**.
+  - Create the source type. 
+      - Click **+ Add Type**.
+      - Enter *source-type* in **Id**.
+      - Select base type *none*.
+      - Add property.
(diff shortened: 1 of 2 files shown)


osisoft-ezrinsky (approved):
Thanks for adding the UID @GiselleLeeOSI !

joyceseng:
Hi Ji Won - thank you for requesting an edit. I have been having a busy week. If I can't get to this soon, I will ask Vicki if she has time. Thank you!

GiselleLeeOSI:
@lasato-OSI and @joyceseng , I'd like to hear your what you think about @osimarkrogers 's suggestion: This procedure needs to be re-written. You do not create new streams as part of the process of creating a stream view. Stream views are designed to be applied to existing streams....if you were creating a new stream, you would simply base it on the type you wanted in the first place. There would be no need to view it through a stream view - - the purpose of a stream view is to enable you to see the archived data of a stream through the lens of a different type. The OCS Portal UI Stream Views feature provides a means to create a stream view, assign two types to it, and define mapping between the properties of the two types (source and target). The Information tab says that to then apply that stream view to an existing stream requires an API call....I am not aware if there is a UI based means of viewing a stream through the stream view. I can take this on to revise in the upcoming sprint if that helps.

I posited a user with no prior knowledge but then it's like writing about how to plant wheat when all you need to provide is how to bake pie. So I'm open to what you think would be a good starting point for the topic.

joyceseng:
I have been having a hell of a week, for various reasons. But I do regard this as my top priority and will get to it as soon as I can. @lasato-OSI if you can look before me, I'd really appreciate it!

lasato-OSI:
> @lasato-OSI and @joyceseng , I'd like to hear your what you think about @osimarkrogers 's suggestion: This procedure needs to be re-written. You do not create new streams as part of the process of creating a stream view. Stream views are designed to be applied to existing streams....if you were creating a new stream, you would simply base it on the type you wanted in the first place. There would be no need to view it through a stream view - - the purpose of a stream view is to enable you to see the archived data of a stream through the lens of a different type. The OCS Portal UI Stream Views feature provides a means to create a stream view, assign two types to it, and define mapping between the properties of the two types (source and target). The Information tab says that to then apply that stream view to an existing stream requires an API call....I am not aware if there is a UI based means of viewing a stream through the stream view. I can take this on to revise in the upcoming sprint if that helps.
> 
> I posited a user with no prior knowledge but then it's like writing about how to plant wheat when all you need to provide is how to bake pie. So I'm open to what you think would be a good starting point for the topic.

(I don't know how to reply in Git Hub.) Anyway, I agree with Mark's approach. There are separate procedures for creating types and streams. I would suggest mentioning that this procedure assumes you've already created your streams views and then point the user to that procedure for more information. By the way, if this doesn't get done before the freeze, let's not worry about it. We can get it in later--either before or after we go live with Zoomin.

osimarkrogers:
> > @lasato-OSI and @joyceseng , I'd like to hear your what you think about @osimarkrogers 's suggestion: This procedure needs to be re-written. You do not create new streams as part of the process of creating a stream view. Stream views are designed to be applied to existing streams....if you were creating a new stream, you would simply base it on the type you wanted in the first place. There would be no need to view it through a stream view - - the purpose of a stream view is to enable you to see the archived data of a stream through the lens of a different type. The OCS Portal UI Stream Views feature provides a means to create a stream view, assign two types to it, and define mapping between the properties of the two types (source and target). The Information tab says that to then apply that stream view to an existing stream requires an API call....I am not aware if there is a UI based means of viewing a stream through the stream view. I can take this on to revise in the upcoming sprint if that helps.
> > I posited a user with no prior knowledge but then it's like writing about how to plant wheat when all you need to provide is how to bake pie. So I'm open to what you think would be a good starting point for the topic.
> 
> (I don't know how to reply in Git Hub.) Anyway, I agree with Mark's approach. There are separate procedures for creating types and streams. I would suggest mentioning that this procedure assumes you've already created your streams views and then point the user to that procedure for more information. By the way, if this doesn't get done before the freeze, let's not worry about it. We can get it in later--either before or after we go live with Zoomin.

That sounds like a good clean solution - - prefacing the Stream Views procedure with the statement that this assumes you have already created streams to work with.  I had also mentioned to Ji Won that even though users will go directly where they want by searching and using links, we should still have these procedures in sequential order.  Dave presented the three tasks in order of Types, Streams, and Stream Views in his Data Organization section in What Does OCS Do...I think if we adhere to the sequence established in that section, then we should be fine with the preface/caveat as well.  But doing that, I would absolutely advocate reducing the Stream Views procedure to just that - - "baking the pie" only.

joyceseng:
Sorry for late response, @GiselleLeeOSI, @osimarkrogers, @lasato-OSI . What i think is this.
I love the introductory sentence and paragraph- yes, I think is totally what is important about stream views. thank you Ji Won for calling this out so clearly at the beginning. 
For the procedure, I'm thinking the user most likely already has the Source stream, but not the Target stream. So I would change this procedure to show that. But I'd add a note saying this procedure assumes you already have the Source stream. 
How does that sound to everyone??

GiselleLeeOSI:
> Sorry for late response, @GiselleLeeOSI, @osimarkrogers, @lasato-OSI . What i think is this.
> I love the introductory sentence and paragraph- yes, I think is totally what is important about stream views. thank you Ji Won for calling this out so clearly at the beginning.
> For the procedure, I'm thinking the user most likely already has the Source stream, but not the Target stream. So I would change this procedure to show that. But I'd add a note saying this procedure assumes you already have the Source stream.
> How does that sound to everyone??

@joyceseng @osimarkrogers and @lasato-OSI , appreciate all your feedback. Now, my question is this: in any of the existing procedures, do you cover how to add data into the streams? It was what stumped me the most when I had to work with stream views myself. Nowhere (including the video, info tab or even API docs) did it mention that you have to have data to see stream views in action. So if that's clearly documented elsewhere, I can absolutely pare down this task to the bare minimum.

joyceseng:
Darn good question, @GiselleLeeOSI. Thanks for bringing this up, that we might not have documented anywhere how to add the data to the stream.

GiselleLeeOSI:
> Darn good question, @GiselleLeeOSI. Thanks for bringing this up, that we might not have documented anywhere how to add the data to the stream.

Okay so if we're all in agreement, I'll absolutely reduce the topic but keep the step where you add data into the streams. To use the analogy we all understand, in PI AF it's no fun creating bunch of elements and attributes and leave it at that (too static). You want to at least hook up your attributes to sinusoid and cdt158 so you see some data in action. Same for stream views, if you want to see the transformation happening, you probably want to work with some dummy data.

osimarkrogers:
> > Darn good question, @GiselleLeeOSI. Thanks for bringing this up, that we might not have documented anywhere how to add the data to the stream.
> 
> Okay so if we're all in agreement, I'll absolutely reduce the topic but keep the step where you add data into the streams. To use the analogy we all understand, in PI AF it's no fun creating bunch of elements and attributes and leave it at that (too static). You want to at least hook up your attributes to sinusoid and cdt158 so you see some data in action. Same for stream views, if you want to see the transformation happening, you probably want to work with some dummy data.
> 
> > Ji Won, this sounds good. When you're done, can you ping us to take a look at it. Thanks.
> > Let's also talk about where to put this topic in the TOC. To address an earlier comment, we can put it after the procedure for creating a stream.

Wow - - this has been a really fruitful discussion!  That sounds like a much more refined approach than we've had so far:
1) Create a type
2) Create a stream
3) Populate the stream with data
4) Create a stream view - with note that it's assumed you've already created a stream (we should probably mention that the value-add of stream views is that you don't lose the archived data of the original stream). Also, I don't believe you need to create a stream based on the new mapped type, either....but we should tell them how to apply the stream view/view the data in the original stream through the new stream view.

GiselleLeeOSI:
> > > Darn good question, @GiselleLeeOSI. Thanks for bringing this up, that we might not have documented anywhere how to add the data to the stream.
> > 
> > 
> > Okay so if we're all in agreement, I'll absolutely reduce the topic but keep the step where you add data into the streams. To use the analogy we all understand, in PI AF it's no fun creating bunch of elements and attributes and leave it at that (too static). You want to at least hook up your attributes to sinusoid and cdt158 so you see some data in action. Same for stream views, if you want to see the transformation happening, you probably want to work with some dummy data.
> > > Ji Won, this sounds good. When you're done, can you ping us to take a look at it. Thanks.
> > > Let's also talk about where to put this topic in the TOC. To address an earlier comment, we can put it after the procedure for creating a stream.
> 
> Wow - - this has been a really fruitful discussion! That sounds like a much more refined approach than we've had so far:
> 
> 1. Create a type
> 2. Create a stream
> 3. Populate the stream with data
> 4. Create a stream view - with note that it's assumed you've already created a stream (we should probably mention that the value-add of stream views is that you don't lose the archived data of the original stream). Also, I don't believe you need to create a stream based on the new mapped type, either....but we should tell them how to apply the stream view/view the data in the original stream through the new stream view.

Okay @osimarkrogers, now I'm really confused because I have in those four steps verbatim in the draft already. I guess what I'm missing is a note about prior knowledge. But you're right about the not needing the target stream if not just for comparison. That can be removed. And good point about the archived data preservation to be added to the prose. Thank you @joyceseng for the "hard"-copy edit lol.

Content_Portal/add-organize-data/configure-stream-views.md
```diff
@@ -0,0 +1,65 @@
+---
+uid: task-configureStreamViews
+---
+
+# Configure stream views
+Stream views provide flexibility in the use of types and allow you to expose data as a different type. 
+While you cannot actually change the properties of types themselves, the stream views feature enables you to create a view of a selected stream that appears as if you had changed the type on which it is based. You create a stream view by choosing a source and target type then a set of mappings between properties of those two types.
```
osimarkrogers, line 7: Recommend "You create a stream view by mapping the properties of the type you want to apply to a stream to the properties of the type with which the stream was originally created. You designate the new type as the source type, and the original type as the target type for the mapping step."

osimarkrogers (changes requested):
Just one editorial suggestion, otherwise I think it covers all the bases.

Content_Portal/add-organize-data/configure-stream-views.md
joyceseng: @osimarkrogers I assume you intend your two sentences to replace the final sentence? The first sentence boggles my mind, too convoluted. But I like the second sentence, as long as it is accurate.

GiselleLeeOSI: It's actually not the type itself that's being mapped but rather the properties that define the type so I think I'm going to keep it as is. But I'll add a bit about preserving archived data by leveraging stream views.

osimarkrogers: The part that caught my eye about yours was the "...choosing a source and target type then a set of mappings..."  That's what I was really trying to remedy.  I wanted to specifically say "mapping the properties of the type you want to apply," but yes that sentence is really convoluted.  My apologies for the stream of consciousness....

osimarkrogers (changes requested):

GiselleLeeOSI:
Updated the topic. 
1. Added information about archived events in the preamble. Marked steps 1-3 optional. 
2. Provided sample data to play with.
3. Removed the step for creating target stream as that is non-essential.
4. Removed the step for adding data into the target stream as that is non-essential.

osimarkrogers (approved):

joyceseng (approved):
Thanks for your thoroughness on this topic, Ji Won.

lasato-OSI (commented):
I may be coming into this discussion belatedly. My apologies. What I was trying to say in an earlier comment is that we already have procedures for creating a type and creating a stream. Do these procedures cover what you describe in steps 1-3? If they do, then I would point users to the procedures rather than repeating the steps here.  One advantage of this is that we avoid having to update these steps in two places. 

It looks like this is more of a tutorial--where you tell them to input specific values to explain a concept like stream views. Is that true? If so, then we should really incorporate it into the Get started topics. It would be very useful to customers.

PR #430, opened 2021-04-05
Updated  what-are-stream-views

joyceseng:
Hi Dave
At last! I edited what-are-stream-views.md. All of my changes were based on trying to make this topic compatible with Ji Won's already existing task topic about configuring stream views.


Content_Portal/terms_staging/what-are-stream-views.md
```diff
@@ -1,15 +1,13 @@
...
-
-OCS provides a graphical interface for setting up stream views, or you can use REST APIs to define stream views programmatically. If you are using the .NET framework, OSIsoft also offers client libraries to help you create and use stream views.
+---
+uid: WhatAreStreamViews
+---
+
+# What are stream views?
+
+<!-- Dave the changes I made here were about making this topic consistent with Ji Won’s topic about configuring stream views. I suggest adding a link to that topic-->
+A stream view is essentially a logical overlay that enables you to view streaming data in a way that is most useful to you. Stream views provide flexibility in the use of types, allowing you to expose data as a different type. While you cannot actually change the properties of types themselves, stream views enable you to create a view of a selected stream that appears as if you had changed the type on which it is based. You create a stream view by choosing a source and target type then a set of mappings between properties of those two types.
```
lasato-OSI, line 8:
```suggestion
A stream view is essentially a logical overlay that enables you to view streaming data in a way that is most useful to you. Stream views provide flexibility in the use of types, allowing you to expose data as a different type. While you cannot actually change the properties of types themselves, stream views enable you to create a view of a selected stream that appears as if you had changed the type on which it is based. You create a stream view by choosing a source and target type, and then creating a set of mappings between the properties of the two types.

```
I have been mystified as to the use case for a stream view--why would you want to create a view of the data with a different type. If it were possible to include an use case, I think it would be helpful.

joyceseng: @lasato-OSI I think the confusion arises because an SDS type is not like a PI Server type. An SDS type is a "composite type" that includes a time stamp and one or more properties. So we are not talking about, for example, changing integer to a string. But about changing the properties that are part of the SDS type. 
This is from the topic About types - we might need to explain this more prominently:
An SDS type is comprised of at least two properties. One property serves as the primary index, most commonly a timestamp or DateTime. In addition, it has one or more additional properties called value properties which describes the data in each stream event. Each value property can have a different property type.

lasato-OSI: Hi Joyce,

Yes, i see that I've had the misunderstanding that you describe. I'm wondering if it might be worthwhile to add one or more of the examples that Derek provided in his email of how stream views are used. What do you think?

laureen

dwicinas-osisoft:
@joyceseng @lasato-OSI I'm finally getting around to reviewing this PR. I agree that it might be worthwhile to explain the differences between types in OCS and types in the PI Data Archive. I don't think this topic is the place for that. Perhaps in the overview on Types? Or better yet, in some kind of section that interprets the differences between OCS and PI Data Archive. I would worry that too much detailed information in these introductory topics might bog down the reader. Users new to OCS would not be interested in that kind of discussion.

PR #438, opened 2021-04-12
Modify SdsTypeCode to return a String

atangOSIsoft:
Asset APIs should return SdsTypeCode as Strings

### DIFF
--- a/Content_Portal/Documentation/Assets/assets.md
+++ b/Content_Portal/Documentation/Assets/assets.md
@@ -35,7 +35,7 @@ An asset or asset type metadata is static information associated with a given as
 | Id          | String | Required*  | `Id` for the metadata value.                                 |
 | Name        | String | Required  | User-friendly name for the metadata value. If not null, must be unique within an asset or asset type. |
 | Description | String | Optional  | User-provided description                                    |
-| SdsTypeCode | Int    | Required  | This integer corresponds to the SdsTypeCode. Asset metadata support the following integer values: 11 (Int64), 14 (Double), 16 (DateTime), and 18 (String). |
+| SdsTypeCode | Int    | Required  | This integer corresponds to the SdsTypeCode. Asset metadata support the following integer or string values: 11 ("Int64"), 14 ("Double"), 16 ("DateTime"), and 18 ("String"). |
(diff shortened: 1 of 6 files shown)


Content_Portal/Documentation/Assets/assets.md
```diff
@@ -35,7 +35,7 @@ An asset or asset type metadata is static information associated with a given as
 | Id          | String | Required*  | `Id` for the metadata value.                                 |
 | Name        | String | Required  | User-friendly name for the metadata value. If not null, must be unique within an asset or asset type. |
 | Description | String | Optional  | User-provided description                                    |
-| SdsTypeCode | Int    | Required  | This integer corresponds to the SdsTypeCode. Asset metadata support the following integer values: 11 (Int64), 14 (Double), 16 (DateTime), and 18 (String). |
+| SdsTypeCode | Int    | Required  | This integer corresponds to the SdsTypeCode. Asset metadata support the following integer or string values: 11 ("Int64"), 14 ("Double"), 16 ("DateTime"), and 18 ("String"). |
```
lasato-OSI, line 38:
```suggestion
| SdsTypeCode | Int    | Required  | This integer corresponds to the SdsTypeCode. Asset metadata support the following integer or string values: 11 ("Int64"), 14 ("Double"), 16 ("DateTime"), and 18 ("String"). |
```
Sorry to bring this nitpicky thing up. I guess I'm not clear on when we're using quotes.

atangOSIsoft: I cannot seem to point out the differences.  I added quotes to indicate the the values in the parenthesis are string literals. Perhaps it's more clear if we separate it such as:
integer values : 11, 14, 16, or 18
corresponding string values: "Int64", "Double", "DateTime", or "String"

GiselleLeeOSI: Perhaps use codeblock instead? `Int64`

lasato-OSI: I think the Content Guild needs to look at this across the API doc and figure out how we're going to indicate string literals and other API elements. There is a "consistency project" and one of its goals is to address issues like this. I suggest we leave it as is and revisit this as part of this project.

PR #500, opened 2021-05-18
Match file name with the toc title

GiselleLeeOSI:
DocuPipe is now updated to match the file name with the h1 header. But the title in the toc still has the old name. Updating the toc once to make sure that they're in line.


digitaljoshua (approved):
Good catch on the small discrepancies. Eagle eyed.

PR #509, opened 2021-05-21
Updated external references (parameters.yml and responses.yml) in prep for next sprint's work

digitaljoshua:
Added a few parameters, added a few response codes

### DIFF
--- a/content/developer-guide/dev-guide-introduction.md
+++ b/content/developer-guide/dev-guide-introduction.md
@@ -1,6 +1,17 @@
 ---
-uid: intro-dev-guide
+uid: lp-dev-guide
 --- 
 # OSIsoft Cloud Services developer guide
 
-OSIsoft Cloud Services developer guide
+OSIsoft Cloud Services (OCS) is a database platform as a service (dbPaaS) designed for real-time operations data.
+Users inside and outside organizations have a flexible, secure, and easy way to access operations data that may be located within a primary control network or at the edge of the industrial network. 
+From data collection to data access to data delivery, OCS provides users with an easy way to prepare and share access to data through a REST API. 
+Contextual information can be added to time-based operations data, enabling process engineers and systems operators to make decisions and take corrective or preemptive actions. 
+Use the information in this guide as you programmatically access OCS. 
+
+## Access to OCS
+
+You must acquire an OCS account before interacting with the OCS. 
+Go to the [OCS page on OSIsoft website](https://www.osisoft.com/pi-system/pi-cloud/osisoft-cloud-services), click [Contact Us](https://www.osisoft.com/contact) and request an OCS account.
+
+
--- a/content/developer-guide/dev-guide-introduction.md
+++ b/content/developer-guide/dev-guide-introduction.md
@@ -3,4 +3,4 @@ uid: intro-dev-guide
 --- 
 # OSIsoft Cloud Services developer guide
 
-OSIsoft Cloud Services developer guide
+OSIsoft Cloud Services developer guide. This is what will appear for users when they click to the developer guide.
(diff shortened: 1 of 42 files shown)


content/developer-guide/dev-guide-introduction.md
```diff
@@ -3,4 +3,4 @@ uid: intro-dev-guide
 --- 
 # OSIsoft Cloud Services developer guide
 
-OSIsoft Cloud Services developer guide
\ No newline at end of file
+OSIsoft Cloud Services developer guide. This is what will appear for users when they click to the developer guide.
```
GiselleLeeOSI, line 6: Hi Josh, this content is no longer valid. See if you can delete this file from your PR. If unsure, talk to me.

content/external-references/parameters.yaml
```diff
@@ -6,18 +6,46 @@ tenantId:
...
+    Client identifier
+id:
+  type: text/plain
+  value: |
+    Unordered list of Ids for all clients to get. Empty or whitespace Ids will be ignored.
+tag:
+  type: text/plain
+  value: |
+    Only return clients that have these tags
 query:
```
GiselleLeeOSI, line 37: query is tricky. There are two kinds: the one that is not supported and the other that is. And the unsupported one is already documented later called query-community. So you shouldn't change the description of this one here.

content/external-references/parameters.yaml
```diff
@@ -6,18 +6,46 @@ tenantId:
...
   type: text/plain
   value: |
-    Query identifier
+    (Not supported) Query identifier
 count:
   type: text/plain
```
GiselleLeeOSI, line 42: Same for count and  and skip. You should keep the original definitions (these are actively being used). Instead you should create count and skip as they apply to identity and label them count-identity(or count-client) and skip-identity(skip-client) respectively.

GiselleLeeOSI (changes requested):
1. Please take a look my comments before making changes. 
2. Please find way to uninclude dev-guide-introduction.md from this pr.

content/developer-guide/dev-guide-introduction.md
digitaljoshua: i deleted this file from the PR (I think).

content/external-references/parameters.yaml
digitaljoshua: reverted back to original definitions; created new query-identity, count-identity, skip-identity

content/external-references/parameters.yaml
digitaljoshua: reverted back to original definitions; created new query-identity, count-identity, skip-identity

digitaljoshua:
not sure if i made a mess of things. 

1. I removed the file content/developer-guide/dev-guide-introduction.md from this PR (hopefully I didn't remove it from the actual branch- I don't think I did but the messages for Git are a bit ambiguous)
2. Changed the modified definition of query back to its original verbiage. I'm sure there is a quick way of selecting that particular change in this PR and poof...cutting that out of the PR. But I just couldn't find it. So I copy/pasted the definition back to its original verbiage that way.
3. Ditto for definition of count
4. Ditto for definition of skip
5. Added 3 new parameters: query-identity, query-count, query-skip

content/external-references/parameters.yaml
```diff
@@ -30,6 +70,34 @@ identityProviderId:
...
+  value: |
+    Number of identity providers to skip
+clientcredentialclientId:
+  type: text/plain
+  value: |
+    Client credential client identifier
+clientcredentialclientCount:
+  type: text/plain
+  value: |
+    Max number of roles to return
```
GiselleLeeOSI, line 96: Between 82 and 96, don't you want to be consistent in your use of Max vs maximum?

digitaljoshua: i fixed this

PR #517, opened 2021-06-03
Added "OCS" to "tenant identifier; added an aadQuery key

digitaljoshua:
Added the qualilfication "OCS" to tenantId description because we deal with an Azure Active Directory tenant.
Added a new key for query for Azure Active Directory

### DIFF
--- a/content/external-references/parameters.yaml
+++ b/content/external-references/parameters.yaml
@@ -5,7 +5,7 @@ namespaceId:
 tenantId:
   type: text/plain
   value: |
-    Tenant identifier
+    OCS tenant identifier
 aadTenantId:
   type: text/plain
   value: |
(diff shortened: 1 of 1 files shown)


content/external-references/parameters.yaml
```diff
@@ -5,7 +5,7 @@ namespaceId:
 tenantId:
   type: text/plain
   value: |
-    Tenant identifier
+    OCS tenant identifier
```
GiselleLeeOSI, line 8: Josh, if you did this it'll change all instances of tenantID to OCS tenant, which would provide unnecessary information in case of SDS, DataViews and pretty much all other routes. I suggest that you create a separate entry to differentiate with AAD tenant and label it tenantId-OCS. 
image

GiselleLeeOSI (changes requested):
Josh, tenantId is used across microservices. Please refrain from making this change.

content/external-references/parameters.yaml
digitaljoshua: I manually reverted the value for tenantId; and added a new key tenantId-OCS

GiselleLeeOSI (approved):

PR #544, opened 2021-06-18
Removed the erroneous internal "TODO: Remove the internal model" comments from 9 API markdown topics

digitaljoshua:
Removed the "TODO: Remove the internal model" sentence from 9 API markdown files.
I will address this in the codebase shortly, but this is to fix in the immediate term on ZoomIn


GiselleLeeOSI (approved):
Looks good!

PR #546, opened 2021-06-22
Added a new key to metadata-tags.yml

digitaljoshua:
Added a new key (bulleted item for namespace) called namespace-identifier-requirements


GiselleLeeOSI (approved):
Looks good!

PR #557, opened 2021-06-24
Updated the first example response body in asset rules api

mskallet:
Updated per pbi-209804: Added AssetTypeId, Status, and example metadata and stream references. Also removed MeasurementMappings.


zmhh:
Am another member on the metadata rules team, this is correct.

PR #558, opened 2021-06-24
Adding Alphanumeric to CharacterType Options

pirogovsky:


content/api-reference/asset-rules/asset-rules.md
```diff
@@ -129,7 +129,7 @@ For more information on available properties for a stream type, see [Sds Types][
...
 | Value              | String     | Optional  | Required only for the iteral and delimiter specification types.<br />Specifies the text and delimiter values. |
 | Name               | String     | Optional  | Required only for the group and wildcard specification types. <br />Specifies the name of the extracted group used in the asset output. Must not begin with a digit and all characters must be a letter, digit or underscore. |
-| CharacterType      | String     | Optional  | Required only for the group specification type.<br />Specified whether the specification allows Any, letters, and digits. If not specified, the default is Any. <!-- Are "Any", "letters", and "digits" the actual values they use? Paul: Yes, that is correct. Only the casing is inconsistent with the code since all are capitalized. The enum is at:L https://dev.azure.com/osieng/engineering/_git/ocs-metadata-rules?path=%2Fsrc%2FOSIsoft.Rules.Models%2FCharacterTypeEnum.cs-->|
+| CharacterType      | String     | Optional  | Required only for the group specification type.<br />Specifies whether the specification allows Any, Letters, Digits or Alphanumeric characters. If not specified, the default is Any. |
```
lasato-OSI, line 132: @pirogovsky Just to clarify, these values are not case-sensitive, are they? That is, would Any, any, and ANY be acceptable?

lasato-OSI, line 132:
```suggestion
| CharacterType      | String     | Optional  | Required only for the group specification type.<br />Specifies the characters allowed in the specification. Valid values are: Any, Letters, Digits or Alphanumeric. If not specified, the default is Any. |
```

lasato-OSI: Confirmed with Mark Skallet that these are case-insensitive values.

PR #574, opened 2021-07-06
Add cache-data-preview parameter to data views external references

jamesbperry:
This PR contributes documentation for a parameter of the Data Views API: `cache`, in the context of the `/preview/dataviews/data/*` routes.

Note that this new external reference (`cache-data-preview`) is an adaptation of an existing external reference (`cache-data`). Both will be used going forward.

### DIFF
--- a/content/external-references/dataviews-parameters.yaml
+++ b/content/external-references/dataviews-parameters.yaml
@@ -20,6 +20,16 @@ cache-data:
     |--|--|
     | `Refresh` | Force the resource to re-resolve.  This is the default value for this API route.  
     | `Preserve`| Use cached information, if available.
+    
+cache-data-preview:
+  type: markdown
+  value: |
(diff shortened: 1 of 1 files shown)


content/external-references/dataviews-parameters.yaml
```diff
@@ -20,6 +20,16 @@ cache-data:
     |--|--|
     | `Refresh` | Force the resource to re-resolve.  This is the default value for this API route.  
     | `Preserve`| Use cached information, if available.
+    
+cache-data-preview:
+  type: markdown
+  value: |
+    Controls when the data view backing resources are to be forcibly refreshed. Used only when requesting the first page of data. Ignored if used with the continuationToken. Values are:
```
osimarkrogers, line 27: Combine middle two sentences: "Used only when requesting the first page of data; ignored if used with the continuationToken."

osimarkrogers (approved):

PR #578, opened 2021-07-07
DocuPipe entries for get allfieldsets

BethZelasky:
Added an example and an entry in the data-views yaml

### DIFF
--- a/content/external-references/dataviews-summaries.yaml
+++ b/content/external-references/dataviews-summaries.yaml
@@ -140,7 +140,7 @@ dataviews-get:
 allfieldsets-get:
   type: text/plain
   value: |
-    Gets the collection of all field sets associated with the data view.
+    Gets the collection of all field sets associated with the data view, including fields which are currently included in the data view.
 
 availablefieldsets-get:
   type: text/plain
--- a/content/external-references/dataviews-summaries.yaml
+++ b/content/external-references/dataviews-summaries.yaml
@@ -137,6 +137,11 @@ dataviews-get:
     Returns a list of data views.
 
 # method summaries: resolved routes
+allfieldsets-get:
+  type: text/plain
+  value: |
+    Gets the collection of all field sets associated with the data view.
+
 availablefieldsets-get:
   type: text/plain
   value: |
--- a/content/external-references/dataviews-summaries.yaml
+++ b/content/external-references/dataviews-summaries.yaml
@@ -140,7 +140,7 @@ dataviews-get:
 allfieldsets-get:
   type: text/plain
   value: |
-    Gets the collection of all field sets associated with the data view, including fields which are currently included in the data view.
+    Gets the collection of all field sets eligible for use in the data view, including fields which are currently included in the data view.
 
 availablefieldsets-get:
   type: text/plain
--- a/content/external-references/dataviews-summaries.yaml
+++ b/content/external-references/dataviews-summaries.yaml
@@ -140,7 +140,7 @@ dataviews-get:
 allfieldsets-get:
   type: text/plain
   value: |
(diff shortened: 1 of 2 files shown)


digitaljoshua (approved):

content/external-references/dataviews-summaries.yaml
```diff
@@ -137,6 +137,11 @@ dataviews-get:
     Returns a list of data views.
 
 # method summaries: resolved routes
+allfieldsets-get:
+  type: text/plain
+  value: |
+    Gets the collection of all field sets associated with the data view, including fields which are currently included in the data view.
```
jamesbperry, line 143: I'm not in love with the wording "associated with." Perhaps "eligible for use in"?

osimarkrogers: I second that.

content/external-references/dataviews-summaries.yaml
```diff
@@ -137,6 +137,11 @@ dataviews-get:
     Returns a list of data views.
 
 # method summaries: resolved routes
+allfieldsets-get:
+  type: text/plain
+  value: |
+    Gets the collection of all field sets eligible for use in the data view, including fields which are currently included in the data view.
```
osimarkrogers, line 143: Change to "Returns the collection...."

osimarkrogers (changes requested):

content/external-references/dataviews-summaries.yaml
BethZelasky: Not intentionally arguing here - but many other lines in the same file start with "Gets the ...." .... so do I change all to "Returns the ..." to keep the symmetry?

BethZelasky: Actually we have a mixture of "Gets ... " and "Returns ..." in that file.  Should they all be "Returns ... " ?

osimarkrogers: Hi Beth,

Yes, sorry I didn't catch those as well.  But yes, "Returns" rather than "Gets" is part of our API doc standard. (I'd include the link, but it's locked up at the moment!)

osimarkrogers (changes requested):

content/external-references/dataviews-summaries.yaml
osimarkrogers: OK, I should clarify: please just fix what's in this PR.  As far as the mountain of other Data Views API doc, we have a separate project running to bring that into compliance.

BethZelasky: Sure - I understand.  Re-pushed.

osimarkrogers (approved):

content/external-references/dataviews-summaries.yaml
```diff
@@ -140,7 +140,7 @@ dataviews-get:
 allfieldsets-get:
   type: text/plain
   value: |
-    Gets the collection of all field sets eligible for use in the data view, including fields which are currently included in the data view.
+    Returns the collection of all field sets eligible for use in the data view, including fields which are currently included in the data view.
```
jamesbperry, line 143: Is "returns" the standard nomenclature we should be using vs "gets"? Seems like our docs are a mixed bag predominated by "gets."

osimarkrogers: Hi Brandon,

That is indeed the case at the moment.  "Returns" is part of the new API Doc standard, so we should implement that in new content going forward.  There is another project underway, OCS Doc Consistency, to bring existing doc into compliance.  Normalizing all of our previous usage of "Gets" to "Returns" is one of the items that should come out of that effort.  I will follow up with Angela Flores and Vicki Touati, the two writers involved with Consistency, to make sure they're aware of this particular issue.

PR #588, opened 2021-07-12
Adding a note on timeout header (Do not merge until Kane says so)

GiselleLeeOSI:
File relocated as a result of restructuring.  Adding a note for after the program release.
@digitaljoshua


digitaljoshua (approved):

osisoft-mbishop:
Followed up with Kane, who confirmed it's safe to merge. Merging.

PR #618, opened 2021-08-03
Aflores ce 240620 markdown syntax

afloresOSI:


VictoriaTouati-AVEVA (approved):
Reviewed IA's edits & applied suggested changes via commits to the Add and organize data & Communities section. Hopefully, I did this correctly!

PR #644, opened 2021-08-16
adding Special Characters in Search Queries section into SDS Search topic

digitaljoshua:
This section also exists in the asset-search-api topic in the API reference.


charlottekwhite:
Completed QA, no suggested changes.

PR #657, opened 2021-09-03
Main lasato asset rule builder

afloresOSI:
You can view a generated version of this content here: https://osisoft-dev.zoominsoftware.io/bundle/ocs-main-lasato-asset-rule-builder/page/add-organize-data/organize-data/assets/create-asset-rule.html


content/add-organize-data/organize-data/assets/work-with-assets.md
```diff
@@ -0,0 +1,15 @@
...
+
+ - If you delete an asset rule, then any assets created with the rule are also deleted. There may be a short delay between the time the rule is deleted and when the related assets are deleted. 
+
+ - If you edit the properties of assets created with asset rules, those edits take precedence over the asset rule configuration. For example, if you edit the asset name, the rule does not overwrite this edit.
+
+ - If you manually delete an asset, the deletion is treated as an edit of the asset and overrides the rule. Therefore, if you delete an asset created by an asset rule, you will not be able to recreate that asset with the same Id. If you inadvertently delete an asset, you can recover the asset by deleting the rule. This will delete any assets created with this rule. Then you can recreate the rule and configure the Id so that it resolves to a different Id. For example, assume that the asset Id in the first rule resolved to `Boiler Tank 1` and this asset is deleted in error. You then create a second rule which the Id resolves to `Region 1 - Boiler Tank 1`. OCS sees this as different asset and will create it.
```
osisoft-mbishop, line 15: Removed some future tense.

```suggestion
 - If you manually delete an asset, the deletion is treated as an edit of the asset and overrides the rule. Therefore, if you delete an asset created by an asset rule, you cannot recreate that asset with the same Id. If you inadvertently delete an asset, you can recover the asset by deleting the rule. This action deletes any assets created with this rule. Then you can recreate the rule and configure the Id so that it resolves to a different Id. For example, assume that the asset Id in the first rule resolves to `Boiler Tank 1` and this asset is deleted in error. You then create a second rule which the Id resolves to `Region 1 - Boiler Tank 1`. OCS sees this as different asset and creates it. 
```

osisoft-mbishop (approved):
Hey Angela, 

Looks good for the most part. What needs immediate action:

* There's a broken xref and a few images that need alt text. I made some inline suggestions to correct them.

Other feedback:

* I think a couple of the topics need a short intro paragraph that offers more context. See inline comments.
* I made a few changes to get rid of past/future tense.
* A made a few other edits for style/brevity, but aren't a big issue.

Nice job!

RealVictorZhang (commented):
I worked with Laureen on the original edition, so that looks good. In addition, Mark's suggestions seem reasonable to me

PR #729, opened 2021-10-25
Updated properties on DataItemField

emonino:
Added new property to DataItemField - StreamId
Removed SummaryType and SummaryDirection properties which were incorrectly added to DataItemField


emonino:
> LGTM
> 
> @emonino , are we wanting to make these changes customer-facing immediately, or do we need to withhold them?

These changes are already in production so this PR can be merged whenever. I'll see if I have permissions to complete it myself.

PR #763, opened 2021-11-17
Rules TOC Reorg

osisoft-mbishop:
* Alphabetized the following TOCs:
    * **API Reference > Rules**
    * **API Reference > Rules > Asset rules**
    * **API Reference > Rules > Metadata rules**
* Updated access control topic titles to "Access control"

[output preview](https://osisoft-dev.zoominsoftware.io/bundle/ocs-docupipe-rules-toc-update/page/api-reference/rules/rules-overview.html)

### DIFF
--- a/content/api-reference/rules/asset-rules/toc.yml
+++ b/content/api-reference/rules/asset-rules/toc.yml
@@ -1,12 +1,12 @@
+- name: Access control
+  topicHref: asset-access-control-list.md
 - name: Asset rules
   topicHref: asset-asset-rules.md
-- name: Asset rules preview
-  topicHref: asset-asset-rule-preview.md
+- name: Asset rules change data 
+  topicHref: asset-asset-change-data.md
 - name: Asset rules errors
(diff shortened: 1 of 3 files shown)


cmanhard (changes requested):
I messed up - "Asset rules change data".  The actual resource is "Asset change data" so the original name was correct.  

Unfortunately, it leaves us with that getting moved up above Asset Rules alphabetically. 

I will discuss this one topic with Gavin, because the first item in the list really should be moved out (Asset Rule Statistics)

osisoft-mbishop:
> I will discuss this one topic with Gavin, because the first item in the list really should be moved out (Asset Rule Statistics)

There's no topic in these TOCs with that title. Did you mean "Asset Rule Suggestion"?

Regardless, I'm going to merge this for now, but we can open a new work item when needed.

osisoft-mbishop:
I did make the title update for "Asset change data" though.

PR #783, opened 2021-12-02
Add strict value mappings to asset rules overview

mskallet:


content/api-reference/rules/asset-rules/asset-rules.md
```diff
@@ -133,6 +134,7 @@ For more information on available properties for a stream type, see [Types][xref
...
 | RequiredDelimiters | Array      | Optional  | Required only for the group specification type.<br />Sequence of required delimiters within a group. If specified, each delimiter within the array must appear in the exact order and number within the group. If not specified, the default value is an empty array. |
 | ValueMappings      | Dictionary | Optional  | Required only for the group and wildcard specification types.<br />A collection of mapped values for each group. If not specified, the default is null. |
+| StrictValueMappings| Boolean    | Optional  | If set to true, for a stream to match a rule, there must be a value / mapping pair present in `ValueMappings`. If false, and a matching value mapping is not present, the value itself will be passed through. For example, if 'pdx' (for Portland) is encountered for the location specification in the rule above, and `StrictValueMappings` is 'false', pdx will be passed through since a ValueMapping for 'pdx' is not found.
```
osisoft-mbishop, line 137: Hi @mskallet , I have a few grammar/style suggestions.

```suggestion
| StrictValueMappings| Boolean    | Optional  | If set to true, for a stream to match a rule, you must set a value / mapping pair within `ValueMappings`. If false, and a matching value mapping is not present, the value itself is passed through. For example, if you set a value mapping of `"PDX":"Portland"` within `ValueMappins` and `StrictValueMappings` is 'false', a value of `PDX` is passed through because the `ValueMappings` object is ignored.
```

osisoft-mbishop: One other question--why are we making this change in GitHub rather than ADO? Ideally we want to make these changes in the code base rather than the docupipe output.

mskallet: I'm fine with the changes you suggested. I would just remove the word "object" since ValueMappings is really a property of an object. 

As far as I know, this file is not created via docupipe. It is manually maintained. We don't have this content in our codebase.

osisoft-mbishop:
Ok, thanks for clarifying.  I’ll make that change and merge it in.

content/api-reference/rules/asset-rules/asset-rules.md
mskallet: @osisoft-mbishop When you make that change, could you please throw this sentence into the description as well? "If not specified, the default value of `true` will be used"

osisoft-mbishop:
Sure. I’ll get that merged here before end of day.

PR #809, opened 2022-01-07
281332 update toc dev guide

osisoft-mbishop:
Hi guys. Can I please get a reivew on my work for ticket [281332](https://dev.azure.com/osieng/engineering/_workitems/edit/281332)?

 I updated the TOC for the Developer's Guide to be alphabetical to match the TOC work we did for the API Reference (which we also alphabetized). I'd like you to confirm that I've indeed alphabetized the Developer's guide TOC, spelled everything correctly, etc.

Here's an [output preview](https://osisoft-dev.zoominsoftware.io/bundle/ocs-281332-update-toc-dev-guide/page/developer-guide/dev-guide-introduction.html).

Exceptions to the rule that *should not* be alphabetized:

* **Developer guide > Samples**. We want Samples to be at the bottom.
* **Developer guide > Data views > Data views stream and asset examples**. Same idea. We want samples at the bottom.
* **Developer guide > Data views > Define a data view > {subtopics}**. These topics are listed in a sequential order and therefore aren't alphabetized.

Thanks! Please ping if you need any clarifications.

### DIFF
--- a/content/developer-guide/asset-dev/toc.yml
+++ b/content/developer-guide/asset-dev/toc.yml
@@ -1,10 +1,10 @@
-- name: Asset and Asset Types
+- name: Asset and asset types
   href: asset-and-asset-types-dev-guide.md
-- name: Asset Search
-  href: asset-search-dev-guide.md
-- name: Asset Status
-  href: asset-status-dev-guide.md
-- name: Resolved Asset
-  href: asset-resolved-dev-guide.md
-- name: Asset Centric Data
+- name: Asset-centric data
   href: asset-centric-dev-guide.md
+- name: Asset search
+  href: asset-search-dev-guide.md
+- name: Asset status
+  href: asset-status-dev-guide.md
+- name: Resolved asset
+  href: asset-resolved-dev-guide.md
--- a/content/developer-guide/omf-dev/toc.yml
+++ b/content/developer-guide/omf-dev/toc.yml
@@ -1,8 +1,8 @@
-- name: OMF ingress specification
-  href: omf-ingress-specification.md
-- name: Topics 
-  href: omf-ingress-topics.md
 - name: OMF editor
   href: omf-editor.md  
+- name: OMF ingress specification
+  href: omf-ingress-specification.md
 - name: Subscriptions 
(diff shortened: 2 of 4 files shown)


jasnames:
Topics under Sequential Data Store aren't alphabetized. I can see that the quick start should be first, but are the rest sequential?

Also, I'm not sure if this is within the scope of this, but there are many disagreements with TOC entries and topic titles:
- Asset and asset types / topic title is just "Assets"
- Asset-centric data / topic title is "Assets Centric Data"
- Asset search / topic title is "Asset Search Criteria"
- Asset status / topic title is "Asset and Asset Type Status"
- Resolved asset / topic title is "Resolved Asset Metadata"
- Recommended workflow - Get data view data / TOC has a dash, topic title has a colon
- Preview data view / topic title is "Preview data view API"
- OMF ingress specification / topic title is "Use OSIsoft Message Format with OSIsoft Cloud Services"
- Quick Start SDS / topic title is "Quick Start: Sequential Data Store" (also start shouldn't be capitalized in TOC)
- Streams write data / topic title is "Write data"
- Streams read data / topic title is "Read data"
- UoM and quantities / topic title is "Units of measure and quantities"

KNyce71:
The only thing that I have to add is Assest-centric data is singular in the TOC but in the topic, it's plural (Assets Centric Data). Again, this also speaks to the TOC/topic disagreements mentioned by @jasnames. 

As long as the disagreements between the TOC and topics are out of scope and the SDS information is sequential (which based on a quick read, it appears to be), I don't see any additional issues.

cmanhard:
We alphabetized the API section - not 100% clear to me that the developer guide should be totally alphabetized.  The former is more like a reference document - so it is easier to find things in specific order - the latter is more like a instruction book - so topics are generally covered in a logical order.

osisoft-mbishop:
Thanks for commenting on this @cmanhard, I should have run this by you before doing the work. I assumed we should make the API Reference/Dev Guide match, but based on what you're saying, I'll likely just kill the PR. Let me run it by @osisoft-rabrego to hear his opinion.

PR #817, opened 2022-01-21
Add asset tag as property

atangOSIsoft:


cmanhard (commented):
It would be nice to also add to the example bodies in https://docs.osisoft.com/bundle/ocs/page/api-reference/assets/assets-assets.html

atangOSIsoft:
> It would be nice to also add to the example bodies in https://docs.osisoft.com/bundle/ocs/page/api-reference/assets/assets-assets.html

Already in a PR to the codebase :) I won't post that PR here since it's internal.

afloresOSI (approved):

PR #825, opened 2022-02-03
Add explanation around null values

TomSeymour:


content/developer-guide/sequential-data-store-dev/read-data/sds-filter-expressions.md
```diff
@@ -27,7 +27,11 @@ The table below lists supported and unsupported types within a filter expression
 | ``UInt16``  |                                 |
 | ``UInt32``   |                                 |
 | ``UInt64``  |                                       |
-The following types are supported for use within a filter expression:
+
+> [!NOTE]
+> While Nullable Types are not supported some supported types can have null values. 
+> Null values encountered when evaluating a filter expression will always return false/no match, 
+> for example, both ``"Value" eq null`` and ``"Value" ne null`` will return false.
```
osisoft-mbishop, line 34: I have a few grammar change requests for this one.

```suggestion
> [!NOTE]
> While Nullable Types are not supported, some supported types can have null values. 
> When evaluating a filter expression, null values always return false/no match. 
> For example, both ``"Value" eq null`` and ``"Value" ne null`` return false.
```

osisoft-mbishop (changes requested):
I have a few grammar nitpicks with this one.

PR #831, opened 2022-02-11
292462 feedback trend page

osisoft-mbishop:
Hey Angela, I made a few changes to this image:

1. I removed the "Tenant" picker because there isn't one anymore.
2. I removed "Notifications" because it didn't seem to follow the theme of controls specifically related to Trend UI. I'm happy to add it back if you disagree.
3. I broke it into sections because I was having trouble numbering it intuitively.


afloresOSI (changes requested):
This is a vast improvement! The grouping really helps clarify the screen elements. The only update needed is that the Trend icon in table row 4A is not rendering in Zoomin.

afloresOSI (approved):

osisoft-mbishop:
@afloresOSI, I noticed that as well and checked in a fix. It's building right now. I'll merge after I confirm it's corrected.

PR #839, opened 2022-03-07
Vtouati pbi 298765 data coercion limits

VictoriaTouati-AVEVA:
Requested update to "Limitations of PI to OCS" and "PI point type change" topics by OldIsNew. Technical review by Rob Schmitz & Janelle Minich. These changes will need to be updated in the ADH repo.


afloresOSI (approved):
Looks good. You can remove our comments if you don't need the history.

PR #842, opened 2022-03-17
fixing some indent issues

osisoft-mbishop:
Fixing some indent issues with notes that were causing note formatting issues and messed up step sequencing. This is a bug I found, not associated with any sprint work.

See problem here: https://osisoft-prod.zoominsoftware.io/bundle/ocs/page/overview/get-started/gs-types.html

Fix (branch output preview): https://osisoft-dev.zoominsoftware.io/bundle/ocs-indent-and-number-sequence-patch/page/overview/get-started/gs-types.html

### DIFF
--- a/content/overview/get-started/gs-roles.md
+++ b/content/overview/get-started/gs-roles.md
@@ -19,11 +19,9 @@ To create a role, follow these steps:
 1. In the `Add Role` window, complete the following fields:
 
    - **Name**  - Enter *MyData.Role*.
-
-
+   
    - **Description** - (Optional) Add a description for this role.
 
-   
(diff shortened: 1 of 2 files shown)


jasnames (approved):
Looks good except for the mix of dashes and en dashes. I'd prefer them to be all en dashes, but at least all one way or the other. (And the \&ndash; HTML isn't really needed when you can just use the – character in markdown.)

VictoriaTouati-AVEVA (approved):
Thanks for fixing the buggy "note" issue. I have battled this bug for awhile & wondered why it pops up sometimes.

PR #849, opened 2022-03-29
updating text per work from 290709

osisoft-mbishop:


osisoft-mbishop:
@vw-pro, this is a doc PR for your work on [290709](https://dev.azure.com/osieng/engineering/_workitems/edit/290709). It was a pretty small doc change--just some content calling out the Sharing tab in SDS lists the communities that you have read access to rather than all communities in the system.

Here's an output preview so you can see the images referenced in the PR: [https://osisoft-dev.zoominsoftware.io/bundle/ocs-294219-sds-explorer-sharing-tab-should-list-all-communities-user-has-read-access-to/page/communities/community-view-shared-streams.html](View shared streams). The applicable content is under the "Sequential Data Store" tab.

afloresOSI (approved):

vw-pro (approved):

PR #853, opened 2022-04-02
Pbi 287191 toc restructure update mb doc qa

osisoft-mbishop:
Hey Jason, here's my doc qa on your toc restructure. For the most part, I thought your changes looked really good. Most of my feedback is title updates for lower-level topics just so that they match the menus in OCS (and it seems like that work might have been out of scope here). Also a few style guide nitpicks.

All of the feedback I made here is granular. If you disagree with any of my feedback, you can check out the branch and revert the individual commit with `git revert <COMMIT#>`. I'll check in with you on Monday to make sure I'm making sense. 

Have a good weekend!

### DIFF
--- a/content/add-organize-data/collect-data/connectors/omf/omf-overview.md
+++ b/content/add-organize-data/collect-data/connectors/omf/omf-overview.md
@@ -2,7 +2,7 @@
 uid: lpomfoverview
 ---
 
-# OMF Connection
+# OMF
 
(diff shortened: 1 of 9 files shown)


afloresOSI (approved):
Mark's edits look spot on.

jasnames (approved):
Well done on the lower menu titles, which I hadn't looked at!

PR #890, opened 2022-06-21
callout updates

jasnames:


osisoft-mbishop:
For the community images where you removed the callouts, we should look at just removing them completely. Without the callouts, those images don't offer much value.

Question--are the purple callouts a graphics CoP thing? I need to look those standards over if so. Either way, they're better than the red we've been using.

osisoft-mbishop (approved):

PR #921, opened 2022-08-12
Victoria fix merge

afloresOSI:
This should revert all the Victoria nonsense from today.


afloresOSI:
This clearly did not revert the changes (https://osisoft-dev.zoominsoftware.io/bundle/ocs-ictoria-fix-merge/page/adh-content-portal-overview.html). At this point I just want to revert back to the last good version from 8-10. Any assistance would be greatly appreaciated.

PR #936, opened 2022-08-19
Quorumsens-DocQA

matthewcharles27:
Doc QA for OCS User docs--Communities


matthewcharles27:
Doc QA general comments:

community-community-roles.md:
For "For example, if you create a new community named `Test Community`, a new community role is added to the tenant named `Test Community Community Member`.", I think it might be clearer to name the example simple "Test" over "test community." It'll be a little easier to read without repeating the word "community" in the example.

community-workflow-create.md
You have a link under Prerequisites (Be assigned [community administration permissions](https://osisoft-dev.zoominsoftware.io/bundle/ocs-quorumsens/page/communities/community-community-roles.html#community-administrators)). Thought on instead, pointing that link here: https://osisoft-dev.zoominsoftware.io/bundle/ocs-quorumsens/page/communities/community-manage-administrators.html? I'm not sure if that would be more useful or more confusing for the reader. This link crops up a couple more times, so depends on what you think about it.

community-create-community.md:
step 3: contact email is not in this menu. Not sure if it was removed.

community-manage-sds-permissions-create.md
Intro: I think it might be clearer to adjust this: "...an administrative user assigned Manage Permissions on the streams must first grant the Tenant Data Steward role (or another role)..." Instead of the parenthetical, we could add a note under the intro to explain that we recommend granting the permissions to the Tenant Data Steward role, but you can give those permissions to other roles instead. (I see it now further down in the doc, but I think up here would be clearer)
for Prerequisites: should we link to instructions on how to check if you have Share permissions?
I would change the "Tips" to a note and just have the link to manage the permissions window, moving the role recommendations up to the intro.

community-accept-invitation.md
I see why you split these up this way, but because of the order of operations when joining a community, I'm not sure this works. It might be better served to join creating and joining a community together to better reflect the order of steps. 
Step 7, we get to looking at the details for the community we were invited to, but it doesn't say how to get to the details. We may want to mimic the instructions in Step 1 of Invite a tenant to a community.
I did NOT see anything about verifying the email address when I clicked the details for the community. I took a screenshot of what I saw - I can show you if you want.
May want to add that you can confirm sharing is active by looking at Data Management -> Communities.

PR #963, opened 2022-10-05
354828 resource usage page updates connect billing

osisoft-mbishop:
Things I did for usage docs:

- Updated license usage information fields.
- swapped daily and monthy everywhere, since we default to daily now.
- updated screencaps
- corrected some text about monthly usage that was outright wrong (stuff about 7 months).


OutofmyOrbit (approved):
Looks good.

PR #974, opened 2022-10-13
Docu pipe client failover

jasnames:


content/api-reference/client-failover/client-failover-groups.md
```diff
@@ -0,0 +1,347 @@
+---
+uid: ""
```
osisoft-mbishop, line 2: It's not a huge issue because the toc references files by relative path rather than UID, but we should notify azurite (I think it's azurite) that their docupipe isn't assigning UIDs to their files. They may need to work with Boson to get docupipe

osisoft-mbishop (approved):
There's an issue with UIDs that we should address, but it can be deferred.

PR #988, opened 2022-11-09
366220 correct incorrect information in search documentation and add more examples

osisoft-mbishop:
Hi @RealVictorZhang ,

I made the changes that we discussed during our call yesterday. I made the following changes:

- Updated the title. The title change jumps this topic to the top of the search results for "search". You can confirm them [here](https://osisoft-dev.zoominsoftware.io/bundle/ocs-366220-merge-two-search-topics/page/ocs-content-portal-overview.html).
- Updated the existing example to accurately describe a tokenized search for "Id:X* AND Metadata/Name:B*"
- Added a new example for "Id:X* AND Metadata/Name:"B"*", which will give the user who raised the ticket the result they were expecting.
- Added the search example from "SDS Search" in the developer documentation.

### DIFF
--- a/content/overview/search.md
+++ b/content/overview/search.md
@@ -47,7 +47,7 @@ Stream metadata keys are only searchable in association with their values. Witho
 
 ## Special characters in search queries
 
-Add the backslash escape character ( \ ) before any special characters in search queries. The following special characters require an escape character: " | / * \ ( ) :
+Add the backslash escape character ( \ ) before any special characters in search queries. The following special characters require an escape character: " | / * \ ( ) : 
 
 The following are examples of using the escape character in query strings.
 
@@ -55,7 +55,7 @@ The following are examples of using the escape character in query strings.
 | -------------------------------------- | ------------------------------------------ |
 | Austin\Dallas\Fort Worth               | Austin\\\Dallas\\\Fort Worth               |
 | 1:100                                  | 1\\:100                                    |
-| http://www.aveva.com                   | http\\:\\/\\/www.aveva.com                 |
+| http://www.aveva.com                   | http\\:\\/\\/www.aveva.com                 | 
 
 ## Examples of query strings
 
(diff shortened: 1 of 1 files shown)


RealVictorZhang (changes requested):
Hi Mark, sorry for getting back late to you on this. I chatted with Gerhard, and we both agreed that this example - Id:X* AND Metadata/Name:B* - could be confusing. I'd recommend we avoid using a Name metadata in examples, and use something industrial operators can relate to:
Id:X* AND Metadata/Location:B*
Returns all items that contain
1. an Id starting with X
2. A metadata wit the name Location
3. and this Location metadata has a value that starts with B such as Boston.

I would suggest we remove the example of Id:MarkedAssetX* AND Metadata/Name:"B"* as I find some unexpected behaviors which I'm working on with R&D. So we will probably need to tweak the documentation again later.
Thanks!

osisoft-mbishop:
@RealVictorZhang , I was out on vacation, and you were probably working on AVEVA world, so no worries about the delay. I updated the example per your comments and removed the second example.

VictorZhang-CAN:
Hi Mark,

I'd suggest we remove the following examples as well, 
- Metadata/Name:*building*
- Metadata/Value:123

or replace them with a real-world example like Metadata/Serieal Number:M0000* which would return all assets that have a metadata with the name Serial Number whose value starts with M0000 (such as M000099 and M000001).

RealVictorZhang:
I'd suggest we remove the following examples as well,
- Metadata/Name:building
- Metadata/Value:123

or replace them with a real-world example like Metadata/Serieal Number:M0000* which would return all assets that have a metadata with the name Serial Number whose value starts with M0000 (such as M000099 and M000001).

RealVictorZhang (commented):
I'd suggest we remove the following examples as well,
- Metadata/Name:building
- Metadata/Value:123

or replace them with a real-world example like Metadata/Serieal Number:M0000* which would return all assets that have a metadata with the name Serial Number whose value starts with M0000 (such as M000099 and M000001).

RealVictorZhang (approved):
This sufficiently addresses the concern raised by the user.

PR #1014, opened 2023-01-13
adding information about autocomplete for namespaces and communities

osisoft-mbishop:
Hey Bo, I made some doc updates for the namespace autocomplete. I also noticed that I was missing autocomplete details about communities, so I added them as well.


osisoft-mbishop:
Should be in ADH repo. Closing.

PR #1027, opened 2023-02-06
382174 create view data documentation

osisoft-mbishop:


osisoft-mbishop:
Preview docs: https://aveva-dev.zoominsoftware.io/bundle/ocs-382174-create-view-data-documentation/page/add-organize-data/store-data/streams/streams-view-data.html

osisoft-mbishop:
@jdspainhower,

These new topics are related to tasks in SDS explorer for:

1. Viewing stream data from SDS explorer.
2. Adding/editing/deleting an event for a data stream.

I haven't written much for SDS explorer, so I'm looking to make sure my terminology is correct and that I generally get the gist of the features. Any and all suggestions welcome!

content/add-organize-data/store-data/streams/streams-add-event.md
```diff
@@ -0,0 +1,29 @@
...
+
+1. Select the stream to add an event for and choose **View Data**.
+
+    **Tip:** Use <xref:Search> to find your stream.
+
+1. Select **Add Event**.
+
+    The **Add Event** panel opens.
+
+1. Select a key property for this stream event (for example, a timestamp).
```
jdspainhower, line 21: I find this step confusing. Are you stating that entering a value for key properties is required?

content/add-organize-data/store-data/streams/streams-view-data.md
```diff
@@ -0,0 +1,45 @@
...
+| Get Range Values | Gets a range of recorded values for the stream. For more information on setting a range, see the heading below.
+
+## Range values
+
+When setting a ranged value query type, you most provide additional parameters to display data values:
+
+| Parameter | Description |
+|--|--|
+| Start Index | The start index for the range. Choose a date and time. By default, this value is the current date and time.  |
+| Count | The number of data values included in the range. By default, this value is 100. |
```
jdspainhower, line 40: I think maybe a better way to phrase this would be: "The maximum number of data values returned by the query. By default, this value is 100."

content/add-organize-data/store-data/streams/streams-view-data.md
```diff
@@ -0,0 +1,45 @@
...
+| Reversed | The order that data values are listed. By default, data values are not reversed. |
```
jdspainhower, line 41: Reversed seems to confuse some users so maybe we could expand on this definition a bit? Essentially, reversed determines whether we query backwards or forwards from the provided Start Index. For example, if reversed=true (and the index is datetime) then we would be searching backwards through time, whereas if reversed=false then we will be searching forward through time from the provided index. I think what throws people is the term "Start Index" because if you are searching datetime values and reversed is true then the Start Index is actually the last value returned chronologically.

content/add-organize-data/store-data/streams/streams-view-data.md
```diff
@@ -0,0 +1,45 @@
...
+
+Select **Apply** to apply your parameters. The data values included in the query are listed on screen, and they are plotted in a graph as well. For streams with multiple properties, you can view different data values included in the stream by editing the **X Axis** and **Y Axis** dropdowns, as depicted in the image below.
```
jdspainhower, line 43: Maybe "Select Apply to apply your parameters and execute the query"?

content/add-organize-data/store-data/streams/streams-add-event.md
osisoft-mbishop: This used to be a step for choosing a timestamp, but @OutofmyOrbit (Robert G) suggested that I use the term "key property," because that property won't necessarily be a timestamp. @OutofmyOrbit, can you give an example of what a key property other than a timestamp may be? This step needs some additional clarification.

content/add-organize-data/store-data/streams/streams-view-data.md
osisoft-mbishop: Updated. Can you take another look to make sure my updates make sense?

content/add-organize-data/store-data/streams/streams-add-event.md
osisoft-mbishop: Updated after talking to Robert. @jdspainhower , can you take another look?

content/add-organize-data/store-data/streams/streams-view-data.md
jdspainhower: Now I think it's confusing in an entirely different way. :)

I think what we are trying to say here is that the property data type will usually be DateTime (the customer could have named that property pretty much anything) but that the data type could be anything (e.g. Double, Integer, etc.).

How about...

The key property is the index value for the event. This property is usually a **DateTime** value, but there is no restriction as to what data type the index can be within SDS.

osisoft-mbishop: I like it. Let's go with that.

jdspainhower: Looks good!

